PR #12, opened 2022-05-10
Wrote erc721 lazyMint contract and test file

jackbonaguro:


contracts/lazymint/ERC721LazyMint.sol
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/// @author: manifold.xyz
+
+import "@manifoldxyz/creator-core-solidity/contracts/core/IERC721CreatorCore.sol";
+import "@manifoldxyz/libraries-solidity/contracts/access/AdminControl.sol";
+
+import "../libraries/ABDKMath64x64.sol";
```
campionfellin, line 10: not used?

contracts/lazymint/ERC721LazyMint.sol
```diff
@@ -0,0 +1,77 @@
...
+    string calldata uri
+  ) external creatorAdminRequired(creatorContractAddress) {
+    // Ensure initialization is only possible once
+    require(!(listings[creatorContractAddress].initialized), "Already initialized");
+
+    // Create the listing!
+    Listing memory listing = Listing({
+      merkleRoot: merkleRoot,
+      uri: uri,
+      initialized: true
```
campionfellin, line 51: In what case is this false? Perhaps like, not use it and maybe like - 

```
    require(!(listings[creatorContractAddress]), "Already initialized");

    // Create the listing!
    Listing memory listing = Listing({
      merkleRoot: merkleRoot,
      uri: uri,
  })
```

contracts/lazymint/ERC721LazyMint.sol
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  function mint(address creatorContractAddress, bytes32[] calldata merkleProof) external {
+      // Verify merkle proof
+      Listing memory listing = listings[creatorContractAddress];
+      bytes32 leaf = keccak256(abi.encodePacked(msg.sender));
+      require(MerkleProof.verify(merkleProof, listing.merkleRoot, leaf), "Could not verify merkle proof");
+
+      // Check for previous mints
+      require(listingTokenEditions[creatorContractAddress][msg.sender] == 0, "Wallet already minted for this contract");
```
campionfellin, line 68: I think you mentioned in discord about customziations, but I wou'dnlt have this as default

contracts/lazymint/ERC721LazyMint.sol
```diff
@@ -0,0 +1,77 @@
...
+
+      // Update listing
+      listingTokenEditions[creatorContractAddress][msg.sender] = 1;
+
+      // Do mint
+      IERC721CreatorCore creatorCoreContract = IERC721CreatorCore(creatorContractAddress);
+      creatorCoreContract.mintExtension(msg.sender, listing.uri);
```
campionfellin, line 75: So this is kinda tricky... I would say _don't_ pass in a token URI here. And make `ERC721LazyMint extend the `ICreatorExtensionTokenURI` interface. Why? So if you mint it like so - there is no way to update the URI after the fact (say, creator has typo or something).

If you mint it with empty and let this "extension" manage the URI, then it _is_ updatable.

contracts/lazymint/ERC721LazyMint.sol
```diff
@@ -0,0 +1,137 @@
...
+    // Create the listing
+    Listing memory newListing = Listing({
+      merkleRoot: merkleRoot,
+      uri: uri,
+      totalMax: totalMax,
+      walletMax: walletMax,
+      startDate: startDate,
+      endDate: endDate
+    });
+    listings[creatorContractAddress][newIndex] = newListing;
```
campionfellin, line 66: Do we need this in memory here? Wonder if we can save some gas on initializing by just like - 

```
    listings[creatorContractAddress][newIndex] = Listing({
      merkleRoot: merkleRoot,
      uri: uri,
      totalMax: totalMax,
      walletMax: walletMax,
      startDate: startDate,
      endDate: endDate
    });;
```

contracts/lazymint/ERC721LazyMint.sol
```diff
@@ -0,0 +1,140 @@
...
+
+  modifier creatorAdminRequired(address creatorContractAddress) {
+    // This is shared, not single-creator. So we must ensure that the initializer is an admin of the contract
+    AdminControl creatorCoreContract = AdminControl(creatorContractAddress);
+    require(creatorCoreContract.isAdmin(msg.sender), "Wallet is not an administrator for contract");
+    _;
+  }
+
+  // Initialize
+  function initializeListing(
```
armellini13, line 42: Wondering if we should have an update function that sets all the parameters again.

In the app, if you edit the draft and publish again, we would then need to have logic to see which parameters changed and use the right functions to update them potentially having more than one transaction (e.g if you update the token and start date you'll need to run 2 txs vs just the initialize function called the first time)

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+
+  modifier creatorAdminRequired(address creatorContractAddress) {
+    // This is shared, not single-creator. So we must ensure that the initializer is an admin of the contract
+    AdminControl creatorCoreContract = AdminControl(creatorContractAddress);
+    require(creatorCoreContract.isAdmin(msg.sender), "Wallet is not an administrator for contract");
+    _;
+  }
+
+  // Initialize
+  function initializeClaim(
```
jtgi, line 41: return `newIndex` pls?

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    // Create the claim
+    claims[creatorContractAddress][newIndex] = Claim({
+      merkleRoot: merkleRoot,
+      uri: uri,
+      totalMax: totalMax,
+      walletMax: walletMax,
+      startDate: startDate,
+      endDate: endDate
+    });
```
armellini13, line 62: @jackbonaguro do we need to return the id of the claim?

contracts/lazyclaim/ERC721LazyClaim.sol
wwhchung: Well, you should emit an event so we can observe.

contracts/lazyclaim/IERC721LazyClaim.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/// @author: manifold.xyz
+
+/**
+ * Lazy Claim interface
+ */
+interface IERC721LazyClaim {
```
wwhchung, line 10: We should pack the struct!  How many are we realistically going to allow ppl to claim?

Dates can be uint48 (meaning that's 96 bits).  Can we pack totalMax, walletMax into the remaining 160 bits?
Also, put the string at the very end.

Meaning the storage will cost 4 slots:
32byte merkleRoot
32byte for totalMax, walletMax, startDate, endDate
arweave URI

But, even better, can we be SUPER smart about things to minimize things:
arweave hash is 43 characters
So, if we store only the arweave tx hash and super compress, that's even better.

contracts/lazyclaim/IERC721LazyClaim.sol
```diff
@@ -0,0 +1,33 @@
...
+    struct Claim {
+        bytes32 merkleRoot;
+        string uri;
+        uint totalMax;
+        uint walletMax;
+        uint startDate;
+        uint endDate;
+    }
+    function initializeClaim(address creatorContractAddress, bytes32 merkleRoot, string calldata uri, uint totalMax, uint walletMax, uint startDate, uint endDate) external;
```
wwhchung, line 19: Change all uints to uint256 or uint (whatever we try to do to supercompress the struct)

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+      if (claim.startDate != 0) require(claim.startDate < block.timestamp, "Transaction before start date");
+      if (claim.endDate != 0) require(claim.endDate >= block.timestamp, "Transaction after end date");
+
+      // Do mint
+      IERC721CreatorCore creatorCoreContract = IERC721CreatorCore(creatorContractAddress);
+      uint newTokenId = creatorCoreContract.mintExtension(msg.sender);
+      tokenClaims[creatorContractAddress][newTokenId] = index;
+  }
+
+  function tokenURI(address creatorContractAddress, uint tokenId) external view returns(string memory) {
```
wwhchung, line 154: We shouldn't do this, we should assume they are all arweave or ipfs, and use the /tokenId format, rather than storing each URI.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+    return claimCounts[creatorContractAddress];
+  }
+  function getClaim(address creatorContractAddress, uint index) external view returns(Claim memory) {
+    return claims[creatorContractAddress][index];
+  }
+
+  // Public mint
+  function mint(address creatorContractAddress, uint index, bytes32[] calldata merkleProof) external {
+      // Verify merkle proof
+      Claim memory claim = claims[creatorContractAddress][index];
```
wwhchung, line 128: Retrieve as storage so we can increment counters directly

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
+import "@openzeppelin/contracts/interfaces/IERC165.sol";
+
+import "./IERC721LazyClaim.sol";
+
+/**
+ * Lazy claim with optional whitelist ERC721 tokens
+ */
+contract ERC721LazyClaim is IERC165, IERC721LazyClaim, ICreatorExtensionTokenURI, ReentrancyGuard {
+  mapping(address => uint) public claimCounts;
```
wwhchung, line 21: Don't need this, store as part of the claim object.

contracts/lazyclaim/IERC721LazyClaim.sol
wwhchung: Acutally, store the current mint count in here too.

contracts/lazyclaim/IERC721LazyClaim.sol
wwhchung: struct ClaimState {
 merkleRoot
 uri
  total;
 totalMax
 walletMax
 startDate
 endDate
 }

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+  mapping(address => mapping(uint => Claim)) public claims;
+  mapping(address => mapping(uint => uint)) public mintsPerClaim;
```
wwhchung, line 23: Where is this configured?  Don't see that being configured, and why not make it part of Claim?

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+  mapping(address => mapping(uint => mapping(address => uint))) public mintsPerWallet;
```
wwhchung, line 24: Structure can be simpler, shouldn't it just be:
index => address => count?

also, plaese document the purpose of each mapping.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+  mapping(address => mapping(uint => uint)) public tokenClaims;
```
wwhchung, line 25: Please name better or do a better description.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+    claims[creatorContractAddress][index] = Claim({
+      merkleRoot: merkleRoot,
+      uri: uri,
+      totalMax: totalMax,
+      walletMax: walletMax,
+      startDate: startDate,
+      endDate: endDate
+    });
+  }
+  function setMerkleRoot(address creatorContractAddress, uint index, bytes32 merkleRoot) external {
```
wwhchung, line 86: Missing permission check? Seems like anyone can set the merkleRoot.  You shuold have permission checks around each function.  My typical strategy to catch this is to look for all external and public functions, and check those.

wwhchung: This is also inefficient.  Just retrieve as storage.  You don't need line 89 if you do.

wwhchung: This goes for all the setX calls btw.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+  }
+  function getClaim(address creatorContractAddress, uint index) external view returns(Claim memory) {
+    return claims[creatorContractAddress][index];
+  }
+
+  // Public mint
+  function mint(address creatorContractAddress, uint index, bytes32[] calldata merkleProof) external {
+      // Verify merkle proof
+      Claim memory claim = claims[creatorContractAddress][index];
+      if (claim.merkleRoot != "") {
```
wwhchung, line 129: Where's the require statement?  Results in unexpected things if you pass in an index that hasn't been made yet.  Need to require index <= currentIndex of creatorContractAddress.

contracts/lazyclaim/ERC721LazyClaim.sol
wwhchung: Better to name this as claimIndex

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+import "@openzeppelin/contracts/interfaces/IERC165.sol";
+
+import "./IERC721LazyClaim.sol";
+
+/**
+ * Lazy claim with optional whitelist ERC721 tokens
+ */
+contract ERC721LazyClaim is IERC165, IERC721LazyClaim, ICreatorExtensionTokenURI, ReentrancyGuard {
+  mapping(address => uint) public claimCounts;
+  mapping(address => mapping(uint => Claim)) public claims;
```
wwhchung, line 22: Comment please

contracts/lazyclaim/ERC721LazyClaim.sol
wwhchung, line 24: Document what the mapping represents.  Like:
// Mapping of number of mints a wallet has done for a claim
// { contractAddress => { claimIndex => { walletAddress => mintCount } } }

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+    claim.startDate = startDate;
+    claims[creatorContractAddress][index] = claim;
+  }
+  function setEndDate(address creatorContractAddress, uint index, uint endDate) external {
+    Claim memory claim = claims[creatorContractAddress][index];
+    claim.endDate = endDate;
+    claims[creatorContractAddress][index] = claim;
+  }
+
+  // Public getters
```
wwhchung, line 117: Don't you need an accessor to get tclaim count for a given index for a given wallet, to show if they can still claim?

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+      Claim memory claim = claims[creatorContractAddress][index];
+      if (claim.merkleRoot != "") {
+        bytes32 leaf = keccak256(abi.encodePacked(msg.sender));
+        require(MerkleProof.verify(merkleProof, claim.merkleRoot, leaf), "Could not verify merkle proof");
+      }
+
+      // Check walletMax against minter's wallet
+      uint walletMinted = mintsPerWallet[creatorContractAddress][index][msg.sender];
+      if (claim.walletMax != 0) require(walletMinted < claim.walletMax, "Maximum tokens already minted for this wallet");
+      mintsPerWallet[creatorContractAddress][index][msg.sender] = walletMinted + 1;
```
wwhchung, line 137: Defer increments until the end (right before mint) just in case there's a further revert).  Do an uncekced ++ rather than +1 to save gas

Basically don't write until all checks are complete

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+
+      // Check totalMax
+      uint claimMinted = mintsPerClaim[creatorContractAddress][index];
```
wwhchung, line 140: No longer need to retrieve from separate mapping because it's part of the claim object.  Similar to prior comment, defer increment till the end.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+      if (claim.totalMax != 0) require(claimMinted < claim.totalMax, "Maximum tokens already minted for this claim");
+      mintsPerClaim[creatorContractAddress][index] = claimMinted + 1;
+
+      // Check timestamps
```
wwhchung, line 144: Do cheap checks first!  And how can it be 0???  You should have a require check as part of this.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+    uint totalMax,
+    uint walletMax,
+    uint startDate,
+    uint endDate
+  ) external creatorAdminRequired(creatorContractAddress) {
+    // Get the index for the new claim
+    uint newIndex = claimCounts[creatorContractAddress];
+    claimCounts[creatorContractAddress] = newIndex + 1;
+
+    // Create the claim
```
wwhchung, line 54: Do input sanity checking first! (require statements)

contracts/lazyclaim/ERC721LazyClaim.sol
wwhchung: Do input sanity checking on each of these.

contracts/lazyclaim/IERC721LazyClaim.sol
wwhchung: In fact, the struct should also have a single byte enum value for 'storageProtocol', enum of
enum StorageProtocol {
  ARWEAVE = 1,
  IPFS = 2
}

And take this as an input.  This way , we can minimize the cost to store the URI.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+      startDate: startDate,
+      endDate: endDate
+    });
+  }
+  function setMerkleRoot(address creatorContractAddress, uint index, bytes32 merkleRoot) external {
+    Claim memory claim = claims[creatorContractAddress][index];
+    claim.merkleRoot = merkleRoot;
+    claims[creatorContractAddress][index] = claim;
+  }
+  function setURI(address creatorContractAddress, uint index, string calldata uri) external {
```
wwhchung, line 91: pass in the storageProtocol (enum byte) and the uri (which is the hash).  Only store that.

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+      mintsPerClaim[creatorContractAddress][index] = claimMinted + 1;
+
+      // Check timestamps
+      if (claim.startDate != 0) require(claim.startDate < block.timestamp, "Transaction before start date");
+      if (claim.endDate != 0) require(claim.endDate >= block.timestamp, "Transaction after end date");
+
+      // Do mint
+      IERC721CreatorCore creatorCoreContract = IERC721CreatorCore(creatorContractAddress);
+      uint newTokenId = creatorCoreContract.mintExtension(msg.sender);
+      tokenClaims[creatorContractAddress][newTokenId] = index;
```
wwhchung, line 151: emit event?

contracts/lazyclaim/ERC721LazyClaim.sol
```diff
@@ -0,0 +1,159 @@
...
+      uint claimMinted = mintsPerClaim[creatorContractAddress][index];
+      if (claim.totalMax != 0) require(claimMinted < claim.totalMax, "Maximum tokens already minted for this claim");
+      mintsPerClaim[creatorContractAddress][index] = claimMinted + 1;
+
+      // Check timestamps
+      if (claim.startDate != 0) require(claim.startDate < block.timestamp, "Transaction before start date");
+      if (claim.endDate != 0) require(claim.endDate >= block.timestamp, "Transaction after end date");
+
+      // Do mint
+      IERC721CreatorCore creatorCoreContract = IERC721CreatorCore(creatorContractAddress);
```
wwhchung, line 149: No need to store variable.  Just do:
IERC721CreatorCore(creatorContractAddress).mintExtension

contracts/lazyclaim/ERC721LazyClaim.sol
wwhchung: Actually, you should apply my IndexRange trick!
https://github.com/manifoldxyz/creator-core-extensions-solidity/blob/9b9039e6d52237a9a3ec847a807de25af8612165/contracts/manifold/edition/ManifoldERC721Edition.sol#L32

In most cases, since ppl won't be minting in the middle of a claim, it will save a TON of gas.

contracts/lazyclaim/ERC721LazyClaim.sol
jtgi: right, an event please :)

jackbonaguro: Added ClaimInitialized & Mint events

contracts/lazyclaim/IERC721LazyClaim.sol
jackbonaguro: Done. Counters are uint32, total size of the stored struct should be 208 bits (including current total counter)

I also left the zero slot of StorageProtocol for NONE in the event someone wants to use their own server

contracts/lazyclaim/IERC721LazyClaim.sol
jackbonaguro: Done

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done with the new `identical` param, otherwise it has no suffix

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Commented all my mappings

Have to push back on naming this `claimIndex` as that name is already used in code to refer to a member of range(0, claimCount)

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Using IndexRanges since most mints will be sequential.

Keeping the name tokenClaims since that's its primary use, but fully documented the mapping

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Removed all the setters in lieu of just the 1 overwrite method with its sanity checks

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done, also done in getters where claim is read

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Commented all mappings; gonna resolve this redundant convo

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Yes you do; implemented `getWalletMinted` method

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Checks now ordered simple->complex. Dates, then totals, then merkle verify.

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done, though only 1 check on initialize (endDate > startDate).

More checks for overwrite, which are also implemented.

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Removed setters but storageProtocol implemented.

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done; indexes cc address, claimIndex, and tokenId. minter address at the end.

contracts/lazyclaim/ERC721LazyClaim.sol
jackbonaguro: Done

jackbonaguro:
Approved!

PR #20, opened 2022-10-11
Simplify lazy claims

jackbonaguro:
- Support EIP 5639
- Add airdrop functionality so creators can mint arbitrary tokens at any time
- Remove restrictions on changing walletMax and totalMax


armellini13:
@jackbonaguro @campionfellin 

Scope creep:

Is there a way to fix the extra token being minted when you create an ERC-1155 claim? Like, can we mint it and burn it in the initialize function for instance? So when creators set a supply of 10 it's actually 10 and not 11.

wwhchung:
It’s an extension. Not sure why you need to mint one first. Should be doable. Shouldn’t need to mint first.

wwhchung:
You could defer that to first claim couldn’t you?

PR #28, opened 2023-01-11
Add ERC721 Burn Redeem shared extension

sullivph:


contracts/manifold/burnredeem/ERC721BurnRedeem.sol
```diff
@@ -0,0 +1,330 @@
...
+        bytes calldata data
+    ) private {
+        // Check calldata is valid
+        require(data.length % 32 == 0, "Invalid data");
+
+        address creatorContractAddress;
+        uint256 burnRedeemIndex;
+        uint256 burnItemIndex;
+        bytes32[] memory merkleProof;
+        (creatorContractAddress, burnRedeemIndex, burnItemIndex, merkleProof) = abi.decode(data, (address, uint256, uint256, bytes32[]));
```
johnnyshankman, line 220: don't we need to encode the group index as well? which group + which item in that group we're burning for?

contracts/manifold/burnredeem/ERC721BurnRedeem.sol
```diff
@@ -0,0 +1,330 @@
...
+
+        _validateBurnRedeem(_burnRedeem);
+        require(
+            _burnRedeem.burnSet.length == 1 &&
+            _burnRedeem.burnSet[0].requiredCount == 1,
+            "Not a 1:1 burn redeem"
+        );
+
+        // Check that the burn token is valid
+        BurnItem memory burnItem = _burnRedeem.burnSet[0].items[burnItemIndex];
```
johnnyshankman, line 232: ah nvm i see this is hardcoded now. this won't work for multiple groups!

contracts/manifold/burnredeem/ERC721BurnRedeem.sol
johnnyshankman: also why is this `bytes32[]`, i was expecting just `bytes32`?

johnnyshankman: also, should this just be 0'd out bytes32 for when the validationType isn't `MERKLE_TREE`?

contracts/manifold/burnredeem/ERC721BurnRedeem.sol
wwhchung: Should fix this Paul or make it more gas efficient.

sullivph: Clarifying that multi-group burn redeems won't work with the onERC721Received transfer hook (see `require` statement on line 225). We are only receiving one token, so anything configuration that requires more than one token to be burned will need to be redeemed with the `burnRedeem` function. How can do we make it more gas efficient?

contracts/manifold/burnredeem/ERC721BurnRedeem.sol
sullivph: > also why is `merkleProof` `bytes32[]`, i was expecting just `bytes32`?

Check out the `verify` function signature here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MerkleProof.sol#L27

sullivph: > also, should this just be 0'd out bytes32 for when the validationType isn't `MERKLE_TREE`?

Ya see unit tests for examples with no merkle proofs

contracts/manifold/burnredeem/ERC721BurnRedeem.sol
wwhchung: Misread. Thought it was being called in the set as well but I see you have validateBurnItem for that.

johnnyshankman: Riiiiiiiiiight only 1155 has the batch receive function. Totally forgot about that.

test/manifold/burnredeem721.js
```diff
@@ -0,0 +1,516 @@
...
+      let balance = await creator.balanceOf(anyone1);
+      assert.equal(0, balance);
+      balance = await burnable721.balanceOf(anyone1);
+      assert.equal(3, balance);
+      balance = await burnable721_2.balanceOf(anyone1);
+      assert.equal(3, balance);
+
+      const merkleElements = [];
+      merkleElements.push(ethers.utils.solidityPack(['uint256'], [3]));
+      merkleElements.push(ethers.utils.solidityPack(['uint256'], [10]));
```
johnnyshankman, line 294: can you try this for me to make sure it's the same as keccak256 is a node-only pacakge:
```
import { keccak256 } from '@ethersproject/solidity';
const leaves = tokenIds.map(
      (x: string) => keccak256(['uint256'], [x])
    );
);
```

I want to ensure the behavior is the same.

johnnyshankman: or actually you could also do this, maybe that's more right:

```
import { keccak256, pack } from '@ethersproject/solidity';
merkleElements.push(pack(['uint256'], [15]));
...
merkleTreeWithValues = new MerkleTree(merkleElements, keccak256, { hashLeaves: true, sortPairs: true });
```

johnnyshankman: So far the only version I can get working on the FE is:

```
    const leaves = tokenIds.map(
      (x: string) => keccak256(['uint256'], [parseInt(x)])
    );
    const tree = new MerkleTree(leaves);
    const root = tree.getRoot().toString('hex');
    ```

johnnyshankman: ugh nevermind it hates ethers version of the method because it requires two params.... blehhhhh what do i do?

johnnyshankman: finally got it, shimmed Buffer 👍

PR #33, opened 2023-02-10
refactor burnredeem

wwhchung:


contracts/manifold/burnredeem/IERC1155BurnRedeem.sol
```diff
@@ -9,24 +9,30 @@ import "./IBurnRedeemCore.sol";
 interface IERC1155BurnRedeem is IBurnRedeemCore {
     struct ExtendedConfig {
```
johnnyshankman, line 10: I think you straight up delete this?

PR #47, opened 2023-03-27
Optimizations

wwhchung:


contracts/manifold/collectible/ERC721Collectible.sol
```diff
@@ -10,308 +10,350 @@ import "@openzeppelin/contracts/utils/Strings.sol";
...
+    /**
+    * @dev override if you want to perform different mint functionality
+    */
+    function _mint(address creatorContractAddress, uint256 instanceId, address to, uint16 amount) internal virtual {
+        CollectibleInstance storage instance = _getInstance(creatorContractAddress, instanceId);
+
+        if (amount == 1) {
+            uint256 tokenId;
+            if (instance.contractVersion >= 3) {
+                uint80 tokenData = uint56(instanceId) << 24 | uint24(++instance.purchaseCount);
```
franjohn21, line 305: It would be nice for code clarity if this was tucked away in a separate function call or similar like -

```
tokenDataPack(uint256 instanceId, uint256 mintOrder): returns uint280 {
  return uint80(uint56(instanceId) << 24 | uint24(mintOrder))
}

tokenDataUnpack(uint80) returns(uint56,uint24) {} 
 ```

File: contracts/manifold/collectible/ERC721Collectible.sol [hunk 2]
```diff
@@ -10,308 +10,350 @@ import "@openzeppelin/contracts/utils/Strings.sol";
...
+                tokenId = IERC721CreatorCore(creatorContractAddress).mintExtension(to, tokenData);
+            } else {
+                ++instance.purchaseCount;
+                tokenId = IERC721CreatorCore(creatorContractAddress).mintExtension(to);
+                _tokenIdToTokenClaimMap[creatorContractAddress][tokenId] = TokenClaim(uint224(instanceId), instance.purchaseCount);
```
franjohn21, line 310: I don't fully understand why we need the `TokenClaim` struct at all now (even for creator core version < 3) -- we could use the same trick for the token claim struct right? I guess the point is that doesn't save us anything since it'd still take up one slot? Still might be cleaner to save it as the same data vs a packed number in one instance and a struct in the other

wwhchung: we can't, because you can mint non-sequentially.

wwhchung: how could we use it properly if things were minted/claimed out of order?

wwhchung: Oh, in version < 3, we didn't have the concept of uint80 token data

contracts/manifold/lazyclaim/ERC721LazyPayableClaim.sol
```diff
@@ -151,6 +162,23 @@ contract ERC721LazyPayableClaim is IERC165, IERC721LazyPayableClaim, ICreatorExt
         return _getClaim(creatorContractAddress, claimIndex);
     }
 
+    /**
+     * See {ILazyPayableClaim-getClaimForToken}.
+     */
+    function getClaimForToken(address creatorContractAddress, uint256 tokenId) public override view returns(Claim memory) {
+        TokenClaim memory tokenClaim = _tokenClaims[creatorContractAddress][tokenId];
+        uint256 claimIndex;
+        if (tokenClaim.claimIndex == 0) {
```
jackbonaguro, line 171: Why not just check the version since this is a view? Seems like a security hole where someone with an "old" contract could put whatever they want in tokenData and reference a claim they don't own

contracts/manifold/lazyclaim/ERC721LazyPayableClaim.sol
```diff
@@ -226,10 +258,19 @@ contract ERC721LazyPayableClaim is IERC165, IERC721LazyPayableClaim, ICreatorExt
         // Transfer funds
         _transferFunds(claim.erc20, claim.cost, claim.paymentReceiver, mintCount, claim.merkleRoot != "");
 
-        uint256[] memory newTokenIds = IERC721CreatorCore(creatorContractAddress).mintExtensionBatch(msg.sender, mintCount);
-        for (uint256 i; i < mintCount;) {
-            _tokenClaims[creatorContractAddress][newTokenIds[i]] = TokenClaim(uint224(claimIndex), uint32(newMintIndex+i));
-            unchecked { ++i; }
+        if (claim.contractVersion >= 3) {
+            uint80[] memory tokenData = new uint80[](mintCount);
```
jackbonaguro, line 262: Can we rename this to `tokenDatas` just to maintain the plural naming convention for arrays?

jackbonaguro: I'm actually not sure if it applies to this line, but there are several places where `tokenData` names an array of tokenDatas

contracts/manifold/lazyclaim/ERC721LazyPayableClaim.sol
```diff
@@ -299,8 +368,19 @@ contract ERC721LazyPayableClaim is IERC165, IERC721LazyPayableClaim, ICreatorExt
      */
     function tokenURI(address creatorContractAddress, uint256 tokenId) external override view returns(string memory uri) {
         TokenClaim memory tokenClaim = _tokenClaims[creatorContractAddress][tokenId];
-        require(tokenClaim.claimIndex > 0, "Token does not exist");
-        Claim memory claim = _claims[creatorContractAddress][tokenClaim.claimIndex];
+        Claim memory claim;
+        uint256 mintOrder;
+        if (tokenClaim.claimIndex != 0) {
```
jackbonaguro, line 373: same thing here, I expected that by incurring the extra upfront cost to store CC version on the claim we would use that as the condition, rather than just having fallback behavior

jackbonaguro (changes requested):

migrations/1_initial_migration.js
```diff
@@ -1,5 +1,13 @@
 const Migrations = artifacts.require("Migrations");
```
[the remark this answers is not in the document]
wwhchung: we need it to link library dependencies (which is now needed since burn redeem had to be split to a new lib due to contract size limits)

contracts/manifold/lazyclaim/ERC721LazyPayableClaim.sol
wwhchung: Renamed

contracts/manifold/collectible/ERC721Collectible.sol
wwhchung: Ya, but trying to reduce gas cost via another indirection/function call.  If you want to address this later, feel free.  Not going in this PR.

contracts/manifold/lazyclaim/ERC721LazyPayableClaim.sol
```diff
@@ -95,199 +105,257 @@ contract ERC721LazyPayableClaim is IERC165, IERC721LazyPayableClaim, ICreatorExt
...
         // Overwrite the existing claim
-        _claims[creatorContractAddress][claimIndex] = Claim({
+        _claims[creatorContractAddress][instanceId] = Claim({
             total: claim.total,
             totalMax: claimParameters.totalMax,
             walletMax: claimParameters.walletMax,
             startDate: claimParameters.startDate,
             endDate: claimParameters.endDate,
             storageProtocol: claimParameters.storageProtocol,
+            contractVersion: claim.contractVersion,
```
campionfellin, line 115: Is this the version of the claim of version of the creator core contract?

contracts/manifold/lazyclaim/ERC721LazyPayableClaim.sol
```diff
@@ -95,199 +105,256 @@ contract ERC721LazyPayableClaim is IERC165, IERC721LazyPayableClaim, ICreatorExt
...
      */
-    function mintBatch(address creatorContractAddress, uint256 claimIndex, uint16 mintCount, uint32[] calldata mintIndices, bytes32[][] calldata merkleProofs, address mintFor) external payable override {
-        Claim storage claim = _getClaim(creatorContractAddress, claimIndex);
+    function mintBatch(address creatorContractAddress, uint256 instanceId, uint16 mintCount, uint32[] calldata mintIndices, bytes32[][] calldata merkleProofs, address mintFor) external payable override {
+        Claim storage claim = _getClaim(creatorContractAddress, instanceId);
 
         // Check totalMax
         claim.total += mintCount;
-        require(claim.totalMax == 0 || claim.total <= claim.totalMax, "Too many requested for this claim");
+        require((claim.totalMax == 0 || claim.total <= claim.totalMax) && claim.total <= MAX_UINT_24, "Too many requested for this claim");
```
jtgi, line 250: claim.total is uint32, why 24?

contracts/manifold/burnredeem/IBurnRedeemCore.sol
```diff
@@ -90,6 +90,7 @@ interface IBurnRedeemCore is IERC165, IERC721Receiver, IERC1155Receiver  {
         uint32 redeemedCount;
         uint16 redeemAmount;
         uint32 totalSupply;
+        uint8 contractVersion;
```
jtgi, line 93: Does this need to be in the middle? dapps will often read structs as arrays, this will shift indices.

contracts/manifold/burnredeem/ERC721BurnRedeem.sol
```diff
@@ -27,63 +30,90 @@ contract ERC721BurnRedeem is BurnRedeemCore, IERC721BurnRedeem {
...
-    function _redeem(address creatorContractAddress, uint256 index, BurnRedeem storage burnRedeemInstance, address to, uint32 count) internal override {
+    function _redeem(address creatorContractAddress, uint256 instanceId, BurnRedeem storage burnRedeemInstance, address to, uint32 count) internal override {
         if (burnRedeemInstance.redeemAmount == 1 && count == 1) {
             ++burnRedeemInstance.redeemedCount;
-            uint256 newTokenId = IERC721CreatorCore(creatorContractAddress).mintExtension(to);
-            _redeemTokens[creatorContractAddress][newTokenId] = RedeemToken(uint224(index), burnRedeemInstance.redeemedCount);
-            emit BurnRedeemMint(creatorContractAddress, index, newTokenId, 1);
+            uint256 newTokenId;
+            if (burnRedeemInstance.contractVersion >= 3) {
+                uint80 tokenData = uint56(instanceId) << 24 | burnRedeemInstance.redeemedCount;
```
jtgi, line 87: where can i learn the purpose of `tokenData`? idempotency token for mints?

`redeemedCount` is `uint32`

contracts/manifold/lazyclaim/ERC721LazyPayableClaim.sol
wwhchung: To fit it in the packed struct nonce (the new creator opttimization).

contracts/manifold/burnredeem/IBurnRedeemCore.sol
wwhchung: ethers will decode it properly if given the abi.  Yes, needs to be in the middle for struct packing (efficiency).  So ya, we need a conditional.

contracts/manifold/burnredeem/ERC721BurnRedeem.sol
wwhchung: no, it doesn't matter it's a uint32.  Will still OR properly and take the lower 32 bits.

tokenData is just immutable data stored with the token when minted that's 'free' due to struct packing (the struct is the mutable ownerAddress and immutable tokenData)

PR #52, opened 2023-05-23
feat: subpackages

campionfellin:
starting work on separation...


johnnyshankman:
Any way to nix the need for `packages/dynamic/test/dynamic/...` and bring those test files up (as a pattern across the repo) so that it's just `packages/dynamic/test/...` and then you can keep the `packages/dynamic/test/helpers...` as is?

Screen Shot 2023-05-30 at 1 26 43 PM

campionfellin:
Ah, good catch on the naming structure. Will fix that up!

PR #58, opened 2023-06-06
Cleanup test organization

Slokh:
Moving tests to be more organized and follow the same directory structure as `contracts/`


Slokh:
> Was the content of any tests changed? Github normally shows it just renamed files, but here is showing them changed?

just prettier formatting, not sure if it wasnt formatted before or my config is different 🤔

PR #71, opened 2023-11-09
add code to do a no burn physical redemption

wwhchung:
Support a no-burn physical redemption


campionfellin:
Merging! @wwhchung

PR #75, opened 2024-01-23
Physicals General Availability Contract

wwhchung:
This implements https://www.notion.so/manifoldxyz/Burn-Hold-Redeem-Physical-Architecture-Redesign-3284301a415c43f88a0f9b2309f39f1e


johnnyshankman (approved):

johnnyshankman:
eh actually who cares, there's no better repo for this unless we make it its OWN repo which is really annoying.

johnnyshankman:
@wwhchung just noticed, does `maxPerWallet` limit enforced anywhere here?

packages/manifold/test/physicalclaim/PhysicalClaimBase.t.sol
```diff
@@ -0,0 +1,37 @@
...
+        // Hack because we were getting stack too deep, so need to pass into subfunction
+        submission = _constructSubmission(instanceId, burnTokens, variation, variationLimit, totalLimit, erc20, price, fundsRecipient, expiration, nonce);
+        submission.instanceId = instanceId;
+    }
+
+    function _constructSubmission(uint256 instanceId, IPhysicalClaim.BurnToken[] memory burnTokens, uint8 variation, uint64 variationLimit, uint64 totalLimit, address erc20, uint256 price, address payable fundsRecipient, uint160 expiration, bytes32 nonce) internal view returns (IPhysicalClaim.BurnSubmission memory submission) {
+        bytes memory messageData = abi.encode(instanceId, burnTokens, variation, variationLimit, totalLimit, erc20, price, fundsRecipient, expiration, nonce);
+        bytes32 message = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", messageData));
```
jackbonaguro, line 19: Is this string correct? I can't find exactly what that \n32 escape character does but I'm suspicious it's not just a newline followed by '32'

johnnyshankman: pretty sure it's some sort of encoded space or new line.

wwhchung: It is 100% correct.  This is how you decode ethereum signed messages and we use it in a number of places.

johnnyshankman:
@wwhchung thought, does maxPerWallet limit need to be enforced anywhere here? that's my only reserve atm, contract looks perfect besides that.

packages/manifold/test/physicalclaim/PhysicalClaimBase.t.sol
wwhchung: Actually let me double check. But the 32 represents the size of data

wwhchung: Doesn't the unit test pass?

wwhchung: Hmmm, 32 might not be the right length here.

johnnyshankman: Found it yes you are correct:
https://eips.ethereum.org/EIPS/eip-191

It's a new line followed by bytes.

> Prompting the user to sign the message with [signMessage()](https://docs.ethers.io/v5/api/signer/#Signer-signMessage) which prefixes the Ethereum ([EIP-191](https://eips.ethereum.org/EIPS/eip-191)) specific identifier "\x19Ethereum Signed Message:\n" and byte length of the message. This

wwhchung: I mean, it 'works' but it's wrong.

wwhchung:
> @wwhchung thought, does maxPerWallet limit need to be enforced anywhere here? that's my only reserve atm, contract looks perfect besides that.

No, because server enforces properly and the 'same' wallet can't do multiple concurrent signed requests as per the server, so no race condition

PR #80, opened 2024-02-08
first pass at new editions contract

campionfellin:
## This PR

This PR creates a new extension contract for minting Editions. This is used by the Series Uploader app (and maybe others). Essentially, it's a gas-efficient way to mint a bunch of tokens to yourself (or others).

The major changes from the old contract is that it takes in an instance ID for idempotency rather than a `seriesIndex`. This is more in-line with moving everything towards instances going forward.

### Screenshots (if applicable)

N/A.

### Ticket

https://linear.app/manifoldxyz/issue/CON-215/[batch-mint]-new-smart-contract

## QA Steps

* [ ]

### DIFF
--- a/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
+++ b/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
@@ -101,6 +101,10 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
         require(recipients.length > 0, "Invalid amount requested");
         require(_totalSupply[creatorCore][instanceId]+recipients.length <= _maxSupply[creatorCore][instanceId], "Too many requested");
         
+        mintTokens(creatorCore, recipients, instanceId);
+    }
+
+    function mintTokens(address creatorCore, address[] memory recipients, uint256 instanceId) internal {
         uint256 startIndex = IERC721CreatorCore(creatorCore).mintExtension(recipients[0]);
         for (uint256 i = 1; i < recipients.length;) {
             IERC721CreatorCore(creatorCore).mintExtension(recipients[i]);
@@ -122,12 +126,7 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
 
         // Mint to recipients
         if (recipients.length > 0) {
-            uint256 startIndex = IERC721CreatorCore(creatorCore).mintExtension(recipients[0]);
-            for (uint256 i = 1; i < recipients.length;) {
-                IERC721CreatorCore(creatorCore).mintExtension(recipients[i]);
-                unchecked{++i;}
-            }
-            _updateIndexRanges(creatorCore, instanceId, startIndex, recipients.length);
+            mintTokens(creatorCore, recipients, instanceId);
         }
 
         return instanceId;
--- a/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
+++ b/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
@@ -159,13 +159,14 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
             uint256 instanceId = _creatorInstanceIds[creatorCore][i];
             IndexRange[] memory indexRanges = _indexRanges[creatorCore][instanceId];
             uint256 offset;
-            for (uint j; j < indexRanges.length; j++) {
+            for (uint j; j < indexRanges.length;) {
                 IndexRange memory currentIndex = indexRanges[j];
                 if (tokenId < currentIndex.startIndex) break;
                 if (tokenId >= currentIndex.startIndex && tokenId < currentIndex.startIndex + currentIndex.count) {
                    return (instanceId, tokenId - currentIndex.startIndex + offset);
                 }
                 offset += currentIndex.count;
+                unchecked{++j;}
             }
             unchecked{++i;}
         }
--- a/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
+++ b/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
@@ -92,17 +92,17 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
         if (_totalSupply[creatorCore][instanceId]+1 > _maxSupply[creatorCore][instanceId]) revert("Too many requested");
         if (recipients.length == 0) revert("No recipients");
 
-        // Grab the start index by minting off a first token...
-        uint256 startIndex = IERC721CreatorCore(creatorCore).mintExtension(recipients[0].recipient);
-        --recipients[0].count; // will revert with underflow if you accidentally have a count of 0... this is good (I think)
-        uint count = 1;
+        uint256 startIndex;
+        uint256 count = 0;
+        uint256[] memory tokenIdResults;
         for (uint256 i; i < recipients.length;) {
+            if (recipients[i].count == 0) revert("Invalid count");
             count += recipients[i].count;
             if (_totalSupply[creatorCore][instanceId]+count > _maxSupply[creatorCore][instanceId]) revert("Too many requested");
-            IERC721CreatorCore(creatorCore).mintExtensionBatch(recipients[i].recipient, recipients[i].count);
+            tokenIdResults = IERC721CreatorCore(creatorCore).mintExtensionBatch(recipients[i].recipient, recipients[i].count);
+            if (i == 0) startIndex = tokenIdResults[0];
             unchecked{++i;}
         }
-
         _updateIndexRanges(creatorCore, instanceId, startIndex, count);
     }
 
--- a/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
+++ b/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
@@ -36,7 +36,7 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
      * @dev Only allows approved admins to call the specified function
      */
     modifier creatorAdminRequired(address creator) {
-        require(IAdminControl(creator).isAdmin(msg.sender), "Must be owner or admin of creator contract");
+        if (!IAdminControl(creator).isAdmin(msg.sender)) revert("Must be owner or admin of creator contract");
         _;
     }
     
@@ -64,7 +64,7 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
      * See {IManifoldERC721Edition-setTokenURIPrefix}.
      */
     function setTokenURIPrefix(address creatorCore, uint256 instanceId, string calldata prefix) external override creatorAdminRequired(creatorCore) {
-        require(_maxSupply[creatorCore][instanceId] != 0, "Invalid instanceId");
+        if (_maxSupply[creatorCore][instanceId] == 0) revert("Invalid instanceId");
         _tokenPrefix[creatorCore][instanceId] = prefix;
     }
     
@@ -80,8 +80,8 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
      * @dev See {IManifoldERC721Edition-mint}.
      */
     function mint(address creatorCore, uint256 instanceId, address recipient, uint16 count) external override nonReentrant creatorAdminRequired(creatorCore) {
-        require(count > 0, "Invalid amount requested");
-        require(_totalSupply[creatorCore][instanceId]+count <= _maxSupply[creatorCore][instanceId], "Too many requested");
+        if (count == 0) revert("Invalid amount requested");
+        if (_totalSupply[creatorCore][instanceId]+count > _maxSupply[creatorCore][instanceId]) revert ("Too many requested");
         
         uint256[] memory tokenIds = IERC721CreatorCore(creatorCore).mintExtensionBatch(recipient, count);
         _updateIndexRanges(creatorCore, instanceId, tokenIds[0], count);
@@ -91,8 +91,8 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
      * @dev See {IManifoldERC721Edition-mint}.
      */
     function mint(address creatorCore, uint256 instanceId, address[] calldata recipients) external override nonReentrant creatorAdminRequired(creatorCore) {
-        require(recipients.length > 0, "Invalid amount requested");
-        require(_totalSupply[creatorCore][instanceId]+recipients.length <= _maxSupply[creatorCore][instanceId], "Too many requested");
+        if (recipients.length == 0) revert("Invalid amount requested");
+        if (_totalSupply[creatorCore][instanceId]+recipients.length > _maxSupply[creatorCore][instanceId]) revert("Too many requested");
         
         mintTokens(creatorCore, recipients, instanceId);
     }
@@ -111,7 +111,7 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
      * @dev See {IManifoldERC721Edition-createSeries}.
      */
     function createSeries(address creatorCore, uint256 maxSupply_, string calldata prefix, uint256 instanceId, address[] memory recipients) external override creatorAdminRequired(creatorCore) returns(uint256) {
-        require(instanceId > 0 && maxSupply_ > 0 && _maxSupply[creatorCore][instanceId] == 0, "Invalid instance");
+        if (instanceId == 0 || maxSupply_ == 0 || _maxSupply[creatorCore][instanceId] != 0) revert("Invalid instance");
         _maxSupply[creatorCore][instanceId] = maxSupply_;
         _tokenPrefix[creatorCore][instanceId] = prefix;
         _creatorInstanceIds[creatorCore].push(instanceId);
--- a/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
+++ b/packages/manifold/contracts/edition/ManifoldERC721Edition.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
 
 /// @author: manifold.xyz
 
-import "@manifoldxyz/libraries-solidity/contracts/access/IAdminControl.sol";
+import "@manifoldxyz/libraries-solidity/contracts/access/AdminControl.sol";
 import "@manifoldxyz/creator-core-solidity/contracts/core/IERC721CreatorCore.sol";
 import "@manifoldxyz/creator-core-solidity/contracts/extensions/CreatorExtension.sol";
 import "@manifoldxyz/creator-core-solidity/contracts/extensions/ICreatorExtensionTokenURI.sol";
@@ -17,7 +17,7 @@ import "./IManifoldERC721Edition.sol";
 /**
  * Manifold ERC721 Edition Controller Implementation
  */
-contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, IManifoldERC721Edition, ReentrancyGuard {
+contract ManifoldERC721Edition is AdminControl, CreatorExtension, ICreatorExtensionTokenURI, IManifoldERC721Edition, ReentrancyGuard {
     using Strings for uint256;
 
     struct IndexRange {
@@ -32,6 +32,10 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
 
     mapping(address => uint256[]) _creatorInstanceIds;
     
+    constructor(address initialOwner) {
+        _transferOwnership(initialOwner);
+    }
+
     /**
      * @dev Only allows approved admins to call the specified function
      */
@@ -40,9 +44,12 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
         _;
     }
     
-    function supportsInterface(bytes4 interfaceId) public view virtual override(CreatorExtension, IERC165) returns (bool) {
-        return interfaceId == type(ICreatorExtensionTokenURI).interfaceId || interfaceId == type(IManifoldERC721Edition).interfaceId ||
-               CreatorExtension.supportsInterface(interfaceId);
+    function supportsInterface(bytes4 interfaceId) public view virtual override(CreatorExtension, IERC165, AdminControl) returns (bool) {
+        return
+            interfaceId == type(ICreatorExtensionTokenURI).interfaceId ||
+            interfaceId == type(IManifoldERC721Edition).interfaceId ||
+            interfaceId == type(IAdminControl).interfaceId ||
+            CreatorExtension.supportsInterface(interfaceId);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,76 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
     }
 
     /**
      * @dev See {IManifoldERC721Edition-mint}.
      */
-    function mint(address creator, uint256 series, address[] calldata recipients) external override nonReentrant creatorAdminRequired(creator) {
+    function mint(address creator, uint256 instanceId, address[] calldata recipients) external override nonReentrant creatorAdminRequired(creator) {
         require(recipients.length > 0, "Invalid amount requested");
-        require(_totalSupply[creator][series]+recipients.length <= _maxSupply[creator][series], "Too many requested");
+        require(_totalSupply[instanceId]+recipients.length <= _maxSupply[instanceId], "Too many requested");
```
wwhchung, line 106: Total supply needs to be creator/instanceId combo as key.  Otherwise ppl could theoretically fake mint by guessing the right instance id's

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -25,11 +25,12 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
         uint256 count;
     }
 
-    mapping(address => mapping(uint256 => string)) _tokenPrefix;
-    mapping(address => mapping(uint256 => uint256)) _maxSupply;
-    mapping(address => mapping(uint256 => uint256)) _totalSupply;
-    mapping(address => mapping(uint256 => IndexRange[])) _indexRanges;
-    mapping(address => uint256) _currentSeries;
+    mapping(uint256 => string) _tokenPrefix;
```
wwhchung, line 28: all these need to key by creator address and instance id otherwise we have a security hole

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -134,22 +128,23 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
     /**
      * @dev Index from tokenId
      */
-    function _tokenSeriesAndIndex(address creator, uint256 tokenId) internal view returns(uint256, uint256) {
-        require(_currentSeries[creator] > 0, "Invalid token");
-        for (uint series=1; series <= _currentSeries[creator]; series++) {
-            IndexRange[] memory indexRanges = _indexRanges[creator][series];
+    function _tokenInstanceAndIndex(address creator, uint256 tokenId) internal view returns(uint256, uint256) {
+        // Go through all their series until we find the tokenId
+        for (uint i = 0; i < _creatorInstanceIds[creator].length; i++) {
```
wwhchung, line 139: uint256 i; i < < _creatorInstanceIds[creator].length;

always use uint256 (be explicit)

Move ++i to unchecked { ++i; } at bottom of loop to save gas.

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: How so - doesn't the `creatorAdminRequired` modifier prevent this?

wwhchung: No it doesn't.  You are keying everything off of instanceId.

So I could instaitate it with MY contract instanceId your future instanceId, and then your frontend app could think it's already instantiated for your creator contract later, because there's no way to differentiate it because your mapping key doesn't contain creator address.

campionfellin: Should be good now

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -134,25 +128,27 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
     /**
      * @dev Index from tokenId
      */
-    function _tokenSeriesAndIndex(address creator, uint256 tokenId) internal view returns(uint256, uint256) {
-        require(_currentSeries[creator] > 0, "Invalid token");
-        for (uint series=1; series <= _currentSeries[creator]; series++) {
-            IndexRange[] memory indexRanges = _indexRanges[creator][series];
+    function _tokenInstanceAndIndex(address creatorCore, uint256 tokenId) internal view returns(uint256, uint256) {
+        // Go through all their series until we find the tokenId
+        for (uint256 i = 0; i < _creatorInstanceIds[creatorCore].length;) {
```
wwhchung, line 139: No need to instantiate i (or j below).  Saves gas

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,76 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
     }
 
     /**
      * See {IManifoldERC721Edition-setTokenURIPrefix}.
      */
-    function setTokenURIPrefix(address creator, uint256 series, string calldata prefix) external override creatorAdminRequired(creator) {
-        require(series > 0 && series <= _currentSeries[creator], "Invalid series");
-        _tokenPrefix[creator][series] = prefix;
+    function setTokenURIPrefix(address creatorCore, uint256 instanceId, string calldata prefix) external override creatorAdminRequired(creatorCore) {
+        require(instanceId > 0, "Invalid instanceId");
```
wwhchung, line 78: Need to check instance is insantiated

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,76 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
         require(recipients.length > 0, "Invalid amount requested");
-        require(_totalSupply[creator][series]+recipients.length <= _maxSupply[creator][series], "Too many requested");
+        require(_totalSupply[creatorCore][instanceId]+recipients.length <= _maxSupply[creatorCore][instanceId], "Too many requested");
         
-        uint256 startIndex = IERC721CreatorCore(creator).mintExtension(recipients[0]);
+        uint256 startIndex = IERC721CreatorCore(creatorCore).mintExtension(recipients[0]);
         for (uint256 i = 1; i < recipients.length;) {
-            IERC721CreatorCore(creator).mintExtension(recipients[i]);
+            IERC721CreatorCore(creatorCore).mintExtension(recipients[i]);
             unchecked{i++;}
```
wwhchung, line 111: ++i is better

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -134,25 +128,27 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
-            IndexRange[] memory indexRanges = _indexRanges[creator][series];
+    function _tokenInstanceAndIndex(address creatorCore, uint256 tokenId) internal view returns(uint256, uint256) {
+        // Go through all their series until we find the tokenId
+        for (uint256 i = 0; i < _creatorInstanceIds[creatorCore].length;) {
+            uint256 instanceId = _creatorInstanceIds[creatorCore][i];
+            IndexRange[] memory indexRanges = _indexRanges[creatorCore][instanceId];
             uint256 offset;
-            for (uint i = 0; i < indexRanges.length; i++) {
-                IndexRange memory currentIndex = indexRanges[i];
+            for (uint j = 0; j < indexRanges.length; j++) {
```
wwhchung, line 143: move j++ to unchecked { ++j; } within loop to save gas

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Done

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Done

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,76 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
-        emit SeriesCreated(msg.sender, creator, series, maxSupply_);
-        return series;
-    }
-
-    /**
-     * @dev See {IManifoldERC721Edition-latestSeries}.
-     */
-    function latestSeries(address creator) external view override returns(uint256) {
-        return _currentSeries[creator];
+    function createSeries(address creatorCore, uint256 maxSupply_, string calldata prefix, uint256 instanceId) external override creatorAdminRequired(creatorCore) returns(uint256) {
```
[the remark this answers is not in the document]
campionfellin: Added

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,98 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
      * @dev See {IManifoldERC721Edition-mint}.
      */
-    function mint(address creator, uint256 series, address[] calldata recipients) external override nonReentrant creatorAdminRequired(creator) {
+    function mint(address creatorCore, uint256 instanceId, address[] calldata recipients) external override nonReentrant creatorAdminRequired(creatorCore) {
         require(recipients.length > 0, "Invalid amount requested");
-        require(_totalSupply[creator][series]+recipients.length <= _maxSupply[creator][series], "Too many requested");
+        require(_totalSupply[creatorCore][instanceId]+recipients.length <= _maxSupply[creatorCore][instanceId], "Too many requested");
         
-        uint256 startIndex = IERC721CreatorCore(creator).mintExtension(recipients[0]);
+        uint256 startIndex = IERC721CreatorCore(creatorCore).mintExtension(recipients[0]);
```
wwhchung, line 106: Factor to helper so we can call the same function within create and mint?

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Done

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,95 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
+     * @dev See {IManifoldERC721Edition-instanceExists}.
      */
-    function createSeries(address creator, uint256 maxSupply_, string calldata prefix) external override creatorAdminRequired(creator) returns(uint256) {
-        _currentSeries[creator] += 1;
-        uint256 series = _currentSeries[creator];
-        _maxSupply[creator][series] = maxSupply_;
-        _tokenPrefix[creator][series] = prefix;
-        emit SeriesCreated(msg.sender, creator, series, maxSupply_);
-        return series;
+    function instanceExists(address creatorCore, uint256 instanceId) external view override returns(bool) {
```
wwhchung, line 59: Eh, could remove this if we have maxSupply function below.  And change maxSupply to revert if zero.  Probably cleaner and less code.

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,95 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
     }
 
     /**
      * See {IManifoldERC721Edition-setTokenURIPrefix}.
      */
-    function setTokenURIPrefix(address creator, uint256 series, string calldata prefix) external override creatorAdminRequired(creator) {
-        require(series > 0 && series <= _currentSeries[creator], "Invalid series");
-        _tokenPrefix[creator][series] = prefix;
+    function setTokenURIPrefix(address creatorCore, uint256 instanceId, string calldata prefix) external override creatorAdminRequired(creatorCore) {
+        require(_maxSupply[creatorCore][instanceId] != 0, "Invalid instanceId");
```
wwhchung, line 74: Let's change all requires to if (x) revert InvalidInstance()

For easier testing?

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -47,83 +48,95 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
+    function mint(address creatorCore, uint256 instanceId, address[] calldata recipients) external override nonReentrant creatorAdminRequired(creatorCore) {
         require(recipients.length > 0, "Invalid amount requested");
-        require(_totalSupply[creator][series]+recipients.length <= _maxSupply[creator][series], "Too many requested");
+        require(_totalSupply[creatorCore][instanceId]+recipients.length <= _maxSupply[creatorCore][instanceId], "Too many requested");
         
-        uint256 startIndex = IERC721CreatorCore(creator).mintExtension(recipients[0]);
+        mintTokens(creatorCore, recipients, instanceId);
+    }
+
+    function mintTokens(address creatorCore, address[] memory recipients, uint256 instanceId) internal {
```
wwhchung, line 107: Naming.  Change to _mintTokens (usually internal or private should have underscore in front).

Why not private? we don't intend to inherit this right?

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: done

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Can do, and I think saves gas

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: done

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -29,101 +29,108 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
+        return string(abi.encodePacked(_tokenPrefix[creatorCore][instanceId], (index+1).toString()));
     }
     
     /**
      * @dev See {IManifoldERC721Edition-mint}.
      */
-    function mint(address creator, uint256 series, address recipient, uint16 count) external override nonReentrant creatorAdminRequired(creator) {
-        require(count > 0, "Invalid amount requested");
-        require(_totalSupply[creator][series]+count <= _maxSupply[creator][series], "Too many requested");
+    function mint(address creatorCore, uint256 instanceId, address recipient, uint16 count) external override nonReentrant creatorAdminRequired(creatorCore) {
```
donpdang, line 84: Why do we need `nonReentrant` here ? Curious cause we don't have this in our other contract mint function such as `LazyClaims721` contract

campionfellin: I believe it's because of the order of the end of the function here...

```
        uint256[] memory tokenIds = IERC721CreatorCore(creatorCore).mintExtensionBatch(recipient, count);
        _updateIndexRanges(creatorCore, instanceId, tokenIds[0], count);
```

If the recipient is a smart contract address, and has `onERC721Received` function that calls back in here, it could continue to loop this function before we call `_updateIndexRanges` to account for the total minted..

@wwhchung might know more

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -29,101 +29,108 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
+            IERC721CreatorCore(creatorCore).mintExtension(recipients[i]);
+            unchecked{++i;}
+        }
+        _updateIndexRanges(creatorCore, instanceId, startIndex, recipients.length);
+    }
+
+    /**
+     * @dev See {IManifoldERC721Edition-createSeries}.
+     */
+    function createSeries(address creatorCore, uint256 maxSupply_, string calldata prefix, uint256 instanceId, address[] memory recipients) external override creatorAdminRequired(creatorCore) returns(uint256) {
```
donpdang, line 114: Does this work if I want to initialize a series and mint a bunch of tokens to my address only?

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -29,101 +29,108 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
     mapping(address => mapping(uint256 => uint256)) _maxSupply;
```
donpdang, line 29: Whats the point of `maxSupply` if the mint function is gated by the creator only?

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
donpdang, line 114: Should check if number of recipients is not more than maxSupply if we keeping the `maxSupply`

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Hmm.. I think it's helpful when they want say 300 tokens, so you can split into 2 transactions. `maxSupply` will be 300 and then `totalSupply` will be 200 after first tx and 300 after the next tx.

But yeah - not sure why we had this in the original contract..

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Yeah good call here.. will do

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
donpdang: ah yea make sense for multi transactions due to gas limit

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Added the checks!

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -29,101 +29,102 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
+        if (_totalSupply[creatorCore][instanceId] != currentSupply) revert("Incorrect supply");
+        _mintTokens(creatorCore, recipients, instanceId);
+    }
+
+    function _mintTokens(address creatorCore, Recipient[] memory recipients, uint256 instanceId) private {
+        if (_totalSupply[creatorCore][instanceId]+1 > _maxSupply[creatorCore][instanceId]) revert("Too many requested");
+        if (recipients.length == 0) revert("No recipients");
+
+        // Grab the start index by minting off a first token...
+        uint256 startIndex = IERC721CreatorCore(creatorCore).mintExtension(recipients[0].recipient);
```
wwhchung, line 96: This is wrong though.  you need to call mintExtensionBatch with the count of the first recipient.  Also you're double minting for the first recipient.

I'm pretty sure mintExtensionBatch will revert with count 0.

The logic is simpler now isn't it?

    uint256[] memory tokenIdResults;
    for (uint256 i; i < recipients.length;) {
           if (count == 0) revert InvalidCount();
            count += recipients[i].count;
            if (_totalSupply[creatorCore][instanceId]+count > _maxSupply[creatorCore][instanceId]) revert TooManyRequested();
            tokenIdResults = IERC721CreatorCore(creatorCore).mintExtensionBatch(recipients[i].recipient, recipients[i].count);
            if (i == 0) startIndex = tokenIdResults[0];
            unchecked{++i;}
        }

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
wwhchung: Ok, with the v3 changes as discussed, this should be changed.

struct EditionInfo {
  uint16 contractVersion;
  uint80 totalSupply;
  uint80 maxSupply;
}

mapping(address => mapping(uint256 => EditionInfo)) _editionInfo;


Now, we can store maxSupply, totalSupply and contract version into a single storage slot, further reducing memory usage.

wwhchung: Also introduce:
    uint256 internal constant MAX_UINT_56 = 0xffffffffffffff;

and ensure instanceId's <= uint56 (for data packing)

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
```diff
@@ -29,101 +29,102 @@ contract ManifoldERC721Edition is CreatorExtension, ICreatorExtensionTokenURI, I
...
-        require(_totalSupply[creator][series]+recipients.length <= _maxSupply[creator][series], "Too many requested");
-        
-        uint256 startIndex = IERC721CreatorCore(creator).mintExtension(recipients[0]);
-        for (uint256 i = 1; i < recipients.length;) {
-            IERC721CreatorCore(creator).mintExtension(recipients[i]);
-            unchecked{i++;}
-        }
-        _updateIndexRanges(creator, series, startIndex, recipients.length);
+    function createSeries(address creatorCore, uint256 maxSupply_, string calldata prefix, uint256 instanceId, Recipient[] memory recipients) external override creatorAdminRequired(creatorCore) returns(uint256) {
+        if (instanceId == 0 || maxSupply_ == 0 || _maxSupply[creatorCore][instanceId] != 0) revert("Invalid instance");
```
wwhchung, line 113: as per above, change to (also no need to return anything anymore) (also argument ordering to be consistent with other extensions, creatorCore and instanceId first, because they are the keys)
```
    function createSeries(address creatorCore, uint256 instanceId, uint24 maxSupply, string calldata prefix, Recipient[] memory recipients) external override creatorAdminRequired(creatorCore) {
    if (instanceId > MAX_UINT_56 || maxSupply == 0 || _editionInfo[creatorCore][instanceId].maxSupply != 0) revert InvalidInput();
    uint8 creatorContractVersion;
        try IERC721CreatorCoreVersion(creatorContractAddress).VERSION() returns(uint256 version) {
            require(version <= 255, "Unsupported contract version");
            creatorContractVersion = uint8(version);
        } catch {}
    _editionInfo[creatorCore][instanceId] = EditionInfo({
         maxSupply: maxSupply,
         tokenPrefix: tokenPrefix,
         contractVersion: creatorContractVersion,
    })
        
 ```

packages/manifold/contracts/edition/ManifoldERC721Edition.sol
campionfellin: Ok, updated to use this

wwhchung: hold on, I'm making some changes

wwhchung: thought of some stuff

PR #82, opened 2024-04-02
feat: add getter function to ManifoldERC721Edition

campionfellin:
## This PR

Adds a getter function to the function so that we can more easily read what editions exist and ensure idempotency when creating new ones.

### Screenshots (if applicable)

N/A


wwhchung:
We don't need this.  Why can't you just call totalSupply?

PR #88, opened 2024-05-24
feat: [CON-944] Gacha extension

nikki-kiga:
## This PR
- adds gacha extension

things left until later:
- max per wallet or max per transaction
- airdrop reserve mints
- updating anything other than start, end, and cost

### Ticket

https://linear.app/manifoldxyz/issue/UPDATE-THIS-PART

## QA Steps

* [ ]


packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,99 @@
...
+    // mvp: Gacha claims cannot be updated except for price
+    event GachaClaimUpdated(address indexed creatorContract, uint256 indexed instanceId);
+    event GachaClaimMintReserved(address indexed creatorContract, uint256 indexed instanceId, address indexed collector, uint256 amount);
+    event GachaClaimMintDelivered(address indexed creatorContract, uint256 indexed instanceId, address indexed collector, uint32 itemIndex, uint256 amount);
+    // nice to have:
+    // event GachaMintBatch(address indexed creatorContract, uint256 indexed instanceId, uint16 mintCount);
+
+    struct Recipient {
+        address receiver;
+        uint256 amount;
```
wwhchung, line 28: count?

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,99 @@
...
+        uint256 payment;
```
wwhchung, line 29: Don't think we need payment

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,99 @@
...
+        uint32 itemIndex;
+    }
+
+    struct Mint {
+        address creatorContractAddress;
+        uint256 instanceId;
+        Recipient[] recipients;
+    }
+        
+    struct ItemProbability {
```
wwhchung, line 39: Don't think we need this.

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,99 @@
...
+        uint32 totalMax;
+        uint48 startDate;
+        uint48 endDate;
+        StorageProtocol storageProtocol;
+        string location;
+        address payable paymentReceiver;
+        address erc20;
+        uint256 cost;
+        uint256 startingTokenId;
+        ItemProbability[] itemProbabilities;
```
wwhchung, line 55: Don't need to store probabilities.

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,99 @@
...
+     * @param creatorContractAddress    the creator contract address
+     * @param instanceId                the claim instanceId for the creator contract
+     * @param mintCount                 the number of claims to mint
+     */
+    function mintReserveBatch(address creatorContractAddress, uint256 instanceId, uint16 mintCount) external payable;
+
+    /**
+     * @notice Deliver NFTs
+     */
+    function deliverMints(Mint[] calldata mints) external;
```
wwhchung, line 87: in deliver, you COULD also keep track of number of things minted for a user and ensure it doesn't go above the number they've requested.

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @notice get mints made for a wallet
+     *
+     * @param minter                    the address of the minting address
+     * @param creatorContractAddress    the address of the creator contract for the claim
+     * @param instanceId                the claim instance for the creator contract
+     * @return                          how many mints the minter has made
+     */
+    function getTotalMints(address minter, address creatorContractAddress, uint256 instanceId) external view returns(uint32);
```
wwhchung, line 97: you actually want to return a struct of:
struct UserMint {
  reserveCount: uint64,
  deliveredCount: uint64,
}

packages/manifold/contracts/gachaclaims/IERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,52 @@
...
+        StorageProtocol storageProtocol;
+        uint32 total;
+        uint32 totalMax;
+        string location;
+        uint48 startDate;
+        uint48 endDate;
+        address payable paymentReceiver;
+        address erc20;
+        uint256 cost;
+        uint256 startingTokenId;
```
wwhchung, line 24: uint112 startingTokenId;
uint8 itemVariations;

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,99 @@
...
+    struct ClaimParameters {
+        uint32 totalMax;
+        uint48 startDate;
+        uint48 endDate;
+        StorageProtocol storageProtocol;
+        string location;
+        address payable paymentReceiver;
+        address erc20;
+        uint256 cost;
+        uint256 startingTokenId;
```
jaxonL, line 54: i dont think we'd have this (it'll be calculated through the implementation of 1155 gacha claim i think..?

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,245 @@
...
+    _validateSigner();
+    _mintReserve(mintReservation);
+  }
+
+  function _mintReserve(MintReservation calldata mintReservation) private {
+    _validateMintReserve(mintReservation);
+    // Checks for reserving
+
+    // Updating the total reserved and mapping
+    // uint256 paymentReceived = msg.value;
```
nikki-kiga, line 162: looking for feedback on how to split logic for checking for payment and minting, and tracking in reservedMints

right now assuming that we don't want to mint until delivery since we need the tokenId

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,245 @@
...
+      revert IGachaLazyClaim.ClaimAlreadyInitialized();
+
+    // Sanity checks
+    if (claimParameters.storageProtocol != StorageProtocol.INVALID) revert IGachaLazyClaim.InvalidStorageProtocol();
+    address[] memory receivers = new address[](1);
+    receivers[0] = msg.sender;
+    string[] memory uris = new string[](1);
+    // TODO check on logic to reserve tokenIds for claim itemVariations
+    uint256[] memory amounts = new uint256[](claimParameters.itemVariations);
+    uint256[] memory newTokenIds = IERC1155CreatorCore(creatorContractAddress).mintExtensionNew(receivers, amounts, uris);
```
nikki-kiga, line 62: right now assuming that this "reserves" tokenIds for the correct number of token variations

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,245 @@
...
+      startingTokenId: uint112(newTokenIds[0]),
+      itemVariations: claimParameters.itemVariations,
+      location: claimParameters.location,
+      paymentReceiver: claimParameters.paymentReceiver,
+      erc20: claimParameters.erc20,
+      cost: claimParameters.cost
+    });
+    for (uint8 i; i < claimParameters.itemVariations; i++) {
+      _claimTokenIds[creatorContractAddress][newTokenIds[i]] = instanceId;
+    }
```
nikki-kiga, line 83: then this maps the related tokenIds for each variant back to the instanceId

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,144 @@
...
+    address indexed collector,
+    uint256 mintCount
+  );
+  event GachaClaimMintDelivered(address indexed creatorContract, uint256 indexed instanceId, address indexed collector);
+  event GachaClaimMintDeliveredBatch(
+    address indexed creatorContract,
+    uint256 indexed instanceId,
+    address indexed collector,
+    uint256 mintCount
+  );
```
nikki-kiga, line 57: are these too granular and/or missing anything that scanner/be would need?

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,245 @@
...
+ */
+contract ERC1155GachaLazyClaim is IERC165, IERC1155GachaLazyClaim, ICreatorExtensionTokenURI, GachaLazyClaim {
+  using Strings for uint256;
+
+  // stores mapping from contractAddress/instanceId to the claim it represents
+  // { contractAddress => { instanceId => Claim } }
+  mapping(address => mapping(uint256 => Claim)) private _claims;
+
+  // { contractAddress => { tokenId => { instanceId } }
+  mapping(address => mapping(uint256 => uint256)) private _claimTokenIds;
```
wwhchung, line 26: Should really rename this to _tokenInstances

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,245 @@
...
+    if (instanceId == 0 || instanceId > MAX_UINT_56) revert IGachaLazyClaim.InvalidInstance();
+    // Revert if claim at instanceId already exists
+    if (_claims[creatorContractAddress][instanceId].storageProtocol == StorageProtocol.INVALID)
+      revert IGachaLazyClaim.ClaimAlreadyInitialized();
+
+    // Sanity checks
+    if (claimParameters.storageProtocol != StorageProtocol.INVALID) revert IGachaLazyClaim.InvalidStorageProtocol();
+    address[] memory receivers = new address[](1);
+    receivers[0] = msg.sender;
+    string[] memory uris = new string[](1);
```
wwhchung, line 59: Make this a length equal to variants so that it actually reserves the correct amount for uris and amounts.  Receivers can be 1 and it should mint to all the same person.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,245 @@
...
+    if (claimParameters.storageProtocol == StorageProtocol.INVALID) revert IGachaLazyClaim.InvalidStorageProtocol();
+    if (keccak256(abi.encodePacked(claimParameters.location)) != keccak256(abi.encodePacked(claim.location))) {
+      revert IGachaLazyClaim.CannotChangeLocation();
+    }
+    if (claimParameters.paymentReceiver != claim.paymentReceiver) revert IGachaLazyClaim.CannotChangePaymentReceiver();
+    if (claimParameters.erc20 != claim.erc20) revert IGachaLazyClaim.CannotChangePaymentToken();
+
+    // Overwrite the existing values for startDate, endDate, and cost
+    _claims[creatorContractAddress][instanceId] = Claim({
+      storageProtocol: claim.storageProtocol,
```
wwhchung, line 111: A lot of these references seeem wrong and should be refrencing claimParameters

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,245 @@
...
+      receivers[i] = mintData.recipients[i].receiver;
+      amounts[i] = mintData.recipients[i].mintCount;
+      unchecked {
+        ++i;
+      }
+    }
+    IERC1155CreatorCore(mintData.creatorContractAddress).mintExtensionExisting(receivers, tokenIds, amounts);
+  }
+
+  // airdrop?
```
wwhchung, line 216: No airdrop

packages/manifold/contracts/gachaclaims/IERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,79 @@
...
+ * Gacha 1155 Lazy Claim interface
+ */
+interface IERC1155GachaLazyClaim is IGachaLazyClaim {
+  struct Claim {
+    StorageProtocol storageProtocol;
+    uint32 total;
+    uint32 totalMax;
+    uint48 startDate;
+    uint48 endDate;
+    uint112 startingTokenId;
```
wwhchung, line 19: Make this a uint80 so it fits in one uint256.

storageProtocol = 8
total = 32
totalMax = 32
startDate = 48
endDate = 48
itemVariations = 8

total = 176
you have 80 left

packages/manifold/contracts/gachaclaims/IERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,79 @@
...
+    uint8 itemVariations;
+    string location;
+    address payable paymentReceiver;
+    address erc20;
+    uint256 cost;
```
wwhchung, line 24: Make cost a uint96, and move it above erc20.  This way you can fit paymentReceiver and cost in one storage slot

packages/manifold/contracts/gachaclaims/IERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,79 @@
...
+    uint48 startDate;
+    uint48 endDate;
+    uint112 startingTokenId;
+    uint8 itemVariations;
+    StorageProtocol storageProtocol;
+    string location;
+    address payable paymentReceiver;
+    address erc20;
+    uint256 cost;
+  }
```
wwhchung, line 38: Fix params her as per above.

packages/manifold/contracts/gachaclaims/IERC1155GachaLazyClaim.sol
wwhchung: address is 160

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * @notice Withdraw funds
+   */
+  function withdraw(address payable receiver, uint256 amount) external;
+
+  /**
+   * @notice                          minting request
+   * @param mintReservation           mint reservation details
+   */
+  function mintReserve(MintReservation calldata mintReservation) external payable;
```
wwhchung, line 101: function mintReserve(address creatorContractAddress, uint256 instanceId, uint32 amount)

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,144 @@
...
+
+  // /**
```
wwhchung, line 103: Remove

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,144 @@
...
+  );
+  event GachaClaimMintDelivered(address indexed creatorContract, uint256 indexed instanceId, address indexed collector);
+  event GachaClaimMintDeliveredBatch(
+    address indexed creatorContract,
+    uint256 indexed instanceId,
+    address indexed collector,
+    uint256 mintCount
+  );
+
+  struct MintReservation {
```
wwhchung, line 59: No need for this struct

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,144 @@
...
+
+  struct Recipient {
+    uint256 mintCount;
+    address receiver;
+  }
+
+  struct Mint {
+    address creatorContractAddress;
+    uint256 instanceId;
+    uint256 tokenId;
```
wwhchung, line 76: uint8 variationIndex

wwhchung: tokenId = startingTokenId + variationIndex

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,144 @@
...
+  //  * @param mintCount                 the number of claims to mint
+  //  */
+  // function mintReserveBatch(address creatorContractAddress, uint256 instanceId, uint16 mintCount) external payable;
+
+  /**
+   * @notice Deliver NFTs
+   */
+  function deliverMints(Mint[] calldata mints) external;
+
+  // add airdrop?
```
wwhchung, line 116: No airdrop.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  /**
+   * See {IGachaLazyClaim-mintReserve}.
+   */
+  function mintReserve(MintReservation calldata mintReservation) external payable override {
+    // _validateEOA(); do we need a validate EOA or validateSigner?
+    _mintReserve(mintReservation);
+  }
+
```
wwhchung, line 156: function mintReserve(address creatorContractAddress, uint256 instanceId, uint32 amount) external payable override {
   Claim storage claim = _getClaim(creatorContractAddress, instanceId);
   // CHECK msg.value == (claim.cost + MINT_FEE)*amount;
   uint32 amountAvailable = claim.totalMax - claim.total;
   if (amountAvailable == 0) revert SoldOut
   uint32 amountToBuy = min(amount, amountAvailable);
   claim.total += amountToBuy;
   // increment user counts by amountToBuy
   // forward value to payment receiver in the amount of amountToBuy*claim.cost
   if (amountToBuy < amount) {
    // refund (amount-amountToBuy)*(claim.cost + MINT_FEE) to sender
   }
}

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
wwhchung: string[] memory uris = new string[](claimParameters.itemVariations);

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,257 @@
...
+    return _getUserMints(minter, creatorContractAddress, instanceId);
+  }
+
+  /**
+   * See {ICreatorExtensionTokenURI-tokenURI}.
+   */
+  function tokenURI(address creatorContractAddress, uint256 tokenId) external view override returns (string memory uri) {
+    uint224 tokenClaim = uint224(_claimTokenIds[creatorContractAddress][tokenId]);
+    if (tokenClaim == 0) revert IGachaLazyClaim.TokenDNE();
+    Claim memory claim = _claims[creatorContractAddress][tokenClaim];
```
wwhchung, line 247: use _getClaim(creatorContractAddress, instanceId) instead

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,257 @@
...
+    uint256 instanceId
+  ) external view override returns (UserMint memory) {
+    return _getUserMints(minter, creatorContractAddress, instanceId);
+  }
+
+  /**
+   * See {ICreatorExtensionTokenURI-tokenURI}.
+   */
+  function tokenURI(address creatorContractAddress, uint256 tokenId) external view override returns (string memory uri) {
+    uint224 tokenClaim = uint224(_claimTokenIds[creatorContractAddress][tokenId]);
```
wwhchung, line 245: rename 'tokenClaim' to instanceId because that's what it is.  No need to cast here.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,257 @@
...
+    if (tokenClaim == 0) revert IGachaLazyClaim.TokenDNE();
+    Claim memory claim = _claims[creatorContractAddress][tokenClaim];
+
+    string memory prefix = "";
+    if (claim.storageProtocol == StorageProtocol.ARWEAVE) {
+      prefix = ARWEAVE_PREFIX;
+    } else if (claim.storageProtocol == StorageProtocol.IPFS) {
+      prefix = IPFS_PREFIX;
+    }
+    uri = string(abi.encodePacked(prefix, claim.location));
```
wwhchung, line 255: string(abi.encodePacked(prefix, claim.location, '/', uint256(tokenId - claim.startingTokenId + 1).toString())

packages/manifold/contracts/gachaclaims/GachaLazyClaim.sol
```diff
@@ -0,0 +1,92 @@
...
+//     if (msg.sender == tx.origin) revert MustUseSignatureMinting();
+//   }
+
+  function _getUserMints(
+    address minter,
+    address creatorContractAddress,
+    uint256 instanceId
+  ) internal view returns (UserMint memory) {
+    UserMint memory userMint = UserMint({
+      receiver: minter,
```
wwhchung, line 86: You don't need receiver.  You already know who it is because it is the minter.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
nikki-kiga: yeah this is intentional for now since we're constraining to only be able to update startDate, endDate and Cost for a gacha claim right now

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
nikki-kiga: ended up modifying the CHECK msg.value == (claim.cost + MINT_FEE)amount; ->     if (msg.value < (claim.cost + MINT_FEE) * mintCount) revert IGachaLazyClaim.InvalidPayment();

as well as to match:
if (amountToBuy < amount) { -> if (msg.value > (claim.cost + MINT_FEE) * amountToReserve){

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+      startDate: claimParameters.startDate,
+      endDate: claimParameters.endDate,
+      startingTokenId: uint80(newTokenIds[0]),
+      itemVariations: claimParameters.itemVariations,
+      location: claimParameters.location,
+      paymentReceiver: claimParameters.paymentReceiver,
+      cost: claimParameters.cost,
+      erc20: claimParameters.erc20
+    });
+    for (uint8 i; i < claimParameters.itemVariations; i++) {
```
wwhchung, line 79: use uint256 for incrementation, it's more gas efficient.  Using anything less than uint256 results in a cost to cast (uint256 is the native machine language construct)

wwhchung: To save gas, you can do an:

unchecked { i++; } within the for loop and get rid of i++.  See other contracts for example

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+   */
+  function updateClaim(
+    address creatorContractAddress,
+    uint256 instanceId,
+    ClaimParameters memory claimParameters
+  ) external override adminRequired {
+    Claim memory claim = _getClaim(creatorContractAddress, instanceId);
+    if (claimParameters.endDate != 0 && claimParameters.startDate >= claimParameters.endDate)
+      revert IGachaLazyClaim.InvalidDate();
+    if (claimParameters.totalMax != claim.totalMax) revert IGachaLazyClaim.CannotChangeTotalMax();
```
wwhchung, line 97: We aren't allowing this change?

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+    if (claimParameters.startingTokenId != claim.startingTokenId) revert IGachaLazyClaim.CannotChangeStartingTokenId();
```
wwhchung, line 98: We should just remove startingTokenId from ClaimParameters.  It should not be part of the struct.

If we use it as a return object, we should create a view only struct.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+    if (claimParameters.itemVariations != claim.itemVariations) revert IGachaLazyClaim.CannotChangeItemVariations();
+    if (claimParameters.storageProtocol != claim.storageProtocol) revert IGachaLazyClaim.CannotChangeStorageProtocol();
```
wwhchung, line 100: Why not? we should allow changing of the storageProtocol, no?  just not setting to Invalid

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+    if (keccak256(abi.encodePacked(claimParameters.location)) != keccak256(abi.encodePacked(claim.location))) {
```
wwhchung, line 101: Why not?

wwhchung: We need to in case there's an asset mistake.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+      revert IGachaLazyClaim.CannotChangeLocation();
+    }
+    if (claimParameters.paymentReceiver != claim.paymentReceiver) revert IGachaLazyClaim.CannotChangePaymentReceiver();
```
wwhchung, line 104: Again, why not?

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
wwhchung: Even if you do that in the app, you should not constrain the smart contract, since it's immutable

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
wwhchung: We should allow but ensure it's not < total

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+  function _getClaim(address creatorContractAddress, uint256 instanceId) private view returns (Claim storage claim) {
+    claim = _claims[creatorContractAddress][instanceId];
+    if (claim.storageProtocol == StorageProtocol.INVALID) revert IGachaLazyClaim.ClaimNotInitialized();
+  }
+
+  /**
+   * See {IGachaLazyClaim-mintReserve}.
+   */
+  function mintReserve(address creatorContractAddress, uint256 instanceId, uint32 mintCount) external payable override {
+    require(!Address.isContract(msg.sender), "Only EOAs can reserve a mint.");
```
wwhchung, line 151: use if and revert, it saves gas.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+    _mintReserve(creatorContractAddress, instanceId, mintCount);
+  }
+
+  function _mintReserve(address creatorContractAddress, uint256 instanceId, uint32 mintCount) private {
+    Claim storage claim = _getClaim(creatorContractAddress, instanceId);
+    // Checks for reserving
+    if (claim.startDate > block.timestamp) revert IGachaLazyClaim.ClaimInactive();
+    if (claim.endDate > 0 && claim.endDate <= block.timestamp) revert IGachaLazyClaim.ClaimInactive();
+    if (msg.value < (claim.cost + MINT_FEE) * mintCount) revert IGachaLazyClaim.InvalidPayment();
```
wwhchung, line 160: Should be using an equality check.  Overpaying is bad.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+    uint32 amountAvailable = claim.totalMax - claim.total;
+    if (amountAvailable == 0 ) revert IGachaLazyClaim.ClaimSoldOut();
+
+    // calculate the amount to reserve and update totals
+    uint32 amountToReserve = uint32(Math.min(mintCount, amountAvailable));
+    claim.total += amountToReserve;
+    _reservedMintsPerWallet[creatorContractAddress][instanceId][msg.sender] += amountToReserve;
+
+    // Refund any overpayment
+    if (msg.value > (claim.cost + MINT_FEE) * amountToReserve){
```
wwhchung, line 170: There is a far cheaper check.  Every time you do additions and multiplications, you increase gas usage
Because we should be enforcing equality of msg.value above, all you need to check here is
if (amountToReserve != mintCount)

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+    }
+    emit GachaClaimMintReserved(creatorContractAddress, instanceId, msg.sender, amountToReserve);
+  }
+
+  /**
+   * See {IGachaLazyClaim-deliverMints}.
+   */
+  function deliverMints(IGachaLazyClaim.Mint[] calldata mints) external override {
+    _validateSigner();
+    _mint(mints);
```
wwhchung, line 182: Move _mint into the same function, every tie you add a sub function it increases gas, and nobody else is using it.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  function _mint(Mint[] calldata mints) private {
+    for (uint256 i; i < mints.length; ) {
+      Mint calldata mintData = mints[i];
+      _mintClaim(mintData);
```
wwhchung, line 188: Combine deliverMints, _mint and _mintClaim.  No reason to separate it all

wwhchung: You are looping twice for no reason, causing a lot of gas usage, because you've split out the functions

packages/manifold/contracts/gachaclaims/GachaLazyClaim.sol
```diff
@@ -0,0 +1,92 @@
...
+  uint256 internal constant MAX_UINT_32 = 0xffffffff;
+  uint256 internal constant MAX_UINT_56 = 0xffffffffffffff;
+  uint256 internal constant MAX_UINT_80 = 0xffffffffffffffffff;
+  uint256 internal constant MAX_UINT_256 = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
+  address internal constant ADDRESS_ZERO = 0x0000000000000000000000000000000000000000;
+
+  uint256 public constant MINT_FEE = 500000000000000;
+
+  // { contractAddress => { instanceId => { walletAddress => reservedMints } } }
+  mapping(address => mapping(uint256 => mapping(address => uint32))) internal _reservedMintsPerWallet;
```
wwhchung, line 30: Only have ONE mapping.  With a single struct called WalletMintInfo

Right now, each person will have two storage slots.  Using a map with a single struct, you can use only one storage slot.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
nikki-kiga: okay updated so everything but paymentToken (consistent with other extensions) and tokenVariations (f.k.a. itemVariations) can be updated

reasoning that tokenVariations would require funky logic and more indexing to handle if other tokens were minted in between gacha claim initialized and updated

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,256 @@
...
+   */
+  function updateClaim(
+    address creatorContractAddress,
+    uint256 instanceId,
+    ClaimParameters memory claimParameters
+  ) external override adminRequired {
+    Claim memory claim = _getClaim(creatorContractAddress, instanceId);
+    if (claimParameters.endDate != 0 && claimParameters.startDate >= claimParameters.endDate)
+      revert IGachaLazyClaim.InvalidDate();
+    if (claimParameters.totalMax < claim.totalMax) revert IGachaLazyClaim.CannotLowerTotalMaxBeyondTotal();
```
wwhchung, line 101: Wrong check.  They CAN lower the totalMax, they just can't lower it below the total minted so far.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,256 @@
...
+    if (claimParameters.tokenVariations < claim.tokenVariations)
```
wwhchung, line 102: Wrong check.  They can't change variations.  It will break the math in tokenURI if those variations were already delivered.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,256 @@
...
+      revert IGachaLazyClaim.CannotLowertokenVariationsBeyondVariations();
+    if (claimParameters.storageProtocol == StorageProtocol.INVALID) revert IGachaLazyClaim.InvalidStorageProtocol();
+    if (claimParameters.erc20 != claim.erc20) revert IGachaLazyClaim.CannotChangePaymentToken();
+    if (claimParameters.tokenVariations != claim.tokenVariations) revert IGachaLazyClaim.CannotChangeTokenVariations();
```
wwhchung, line 106: Wait, this is a dupe.  You can remove the above.

wwhchung: Wait, just change the ClaimParameters or create a new UpdateClaimParameters struct so you don't need these checks...

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,256 @@
...
+    if (claim.endDate > 0 && claim.endDate <= block.timestamp) revert IGachaLazyClaim.ClaimInactive();
+    if (msg.value != (claim.cost + MINT_FEE) * mintCount) revert IGachaLazyClaim.InvalidPayment();
+    uint32 amountAvailable = claim.totalMax - claim.total;
+    if (amountAvailable == 0) revert IGachaLazyClaim.ClaimSoldOut();
+
+    // calculate the amount to reserve and update totals
+    uint32 amountToReserve = uint32(Math.min(mintCount, amountAvailable));
+    claim.total += amountToReserve;
+    _mintDetailsPerWallet[creatorContractAddress][instanceId][msg.sender].reservedCount += amountToReserve;
+
```
wwhchung, line 167: Um, you forgot to pay the creator lol

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,105 @@
...
+    uint256 indexed instanceId,
+    address indexed collector,
+    uint32 mintCount
+  );
+
+  struct Recipient {
+    uint256 mintCount;
+    address receiver;
+  }
+
```
wwhchung, line 53: We need a better structure so we aren't repeatedly looping and retrieving a claim unecessarily.
The structure should really be:
struct ClaimMint {
  address creatorContractAddress;
  uint256 instanceId;
  VariationMint[] variationMint;
}

struct VariationMint {
  uint8 variationIndex;
  Recipient[] recipients;
}

This way you can do one outer for loop and retrieve the Claim only once, and do one variation mint per variation to all recipients.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,256 @@
...
+      Mint calldata mintData = mints[i];
+      Claim memory claim = _getClaim(mintData.creatorContractAddress, mintData.instanceId);
+      uint256[] memory tokenIds = new uint256[](1);
+      if (mintData.variationIndex > claim.tokenVariations || mintData.variationIndex < 1)
+        revert IGachaLazyClaim.InvalidVariationIndex();
+      tokenIds[0] = claim.startingTokenId + mintData.variationIndex - 1;
+      address[] memory receivers = new address[](mintData.recipients.length);
+      uint256[] memory amounts = new uint256[](mintData.recipients.length);
+      for (uint256 j; j < mintData.recipients.length; ) {
+        if (
```
wwhchung, line 191: Reference accesses to maps are expensive.  If you are using/accessing more than once, you want to retrieve it into a storage variable.

MintDetails memory receiverMintDetails = _mintDetailsPerWallet[mintData.creatorContractAddress]

etc.

But this whole loop will likely change as per the below comments.

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
nikki-kiga: Wait am I missing something or will this require 3 loops? Are we no longer accepting an array of mints / claimMints?

Agree that this will save on loops required to call mint function but we would still need to loop through all the recipients to check against reserved and update delivered right?

nikki-kiga: and actually when i was fixing the references i found some bugs with the order of how things were getting incremented which might contributing to above ^^

wwhchung: Yes. You still need to loop recipients to update counts.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
wwhchung: @nikki-kiga we should be able to update the storage protocol.  Right here you seem to be unable to in line 106

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,254 @@
...
+      for (uint256 j; j < mintData.recipients.length; ) {
+        address receiver = mintData.recipients[j].receiver;
+        uint256 mintCount = mintData.recipients[j].mintCount;
+        UserMint storage userDetails = _mintDetailsPerWallet[mintData.creatorContractAddress][mintData.instanceId][receiver];
+
+        if (userDetails.deliveredCount + mintCount > userDetails.reservedCount)
+          revert IGachaLazyClaim.CannotMintMoreThanReserved();
+        receivers[j] = receiver;
+        amounts[j] = mintCount;
+        _mintDetailsPerWallet[mintData.creatorContractAddress][mintData.instanceId][receiver].deliveredCount+= uint32(mintCount);
```
wwhchung, line 199: change this to read:
userDetails.deliveredCount += mintCount

Can reference the storage variable to save gas.  Should not need to cast mintCount since it's already a uint32.  (if it's not change it so it is)

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
wwhchung: Wait, why not? I can think of how the cron can easily figure this out in the mint script.  @jaxonL

wwhchung: @nikki-kiga @jaxonL just changed it.  Don't need Recipient struct.

Why?  Because the external mint call can mint multiple tokens to multiple receivers in multiple amounts in one shot.

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,256 @@
...
+   */
+  function mintReserve(address creatorContractAddress, uint256 instanceId, uint32 mintCount) external payable override {
+    if (Address.isContract(msg.sender)) revert IGachaLazyClaim.CannotMintFromContract();
+    Claim storage claim = _getClaim(creatorContractAddress, instanceId);
+    // Checks for reserving
+    if (claim.startDate > block.timestamp) revert IGachaLazyClaim.ClaimInactive();
+    if (claim.endDate > 0 && claim.endDate <= block.timestamp) revert IGachaLazyClaim.ClaimInactive();
+    if (msg.value != (claim.cost + MINT_FEE) * mintCount) revert IGachaLazyClaim.InvalidPayment();
+    uint32 amountAvailable = claim.totalMax - claim.total;
+    if (amountAvailable == 0) revert IGachaLazyClaim.ClaimSoldOut();
```
jaxonL, line 157: ran into an edge case here! if claim.totalMax is 0, and on a first mint, where claim.total = 0, the amount available is still 0 (and for claim.totalMax = 0, it might underflow in future mints)

logic here should prolly be

if (claim.totalMax != 0) // check available amounts?

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,256 @@
...
+
```
jaxonL, line 158: for sanity, should we revert if mintCount = 0 for some reason?

packages/manifold/test/gacha/ERC1155GachaLazyClaim.t.sol
```diff
@@ -0,0 +1,583 @@
...
+    // update tokenURI
+    example.updateTokenURIParams(address(creatorCore1), 1, IGachaLazyClaim.StorageProtocol.ARWEAVE, "arweaveHashNEW");
+    assertEq("https://arweave.net/arweaveHashNEW/1", creatorCore1.uri(1));
+    assertEq("https://arweave.net/arweaveHashNEW/2", creatorCore1.uri(2));
+    assertEq("https://arweave.net/arweaveHashNEW/3", creatorCore1.uri(3));
+    assertEq("https://arweave.net/arweaveHashNEW/4", creatorCore1.uri(4));
+    assertEq("https://arweave.net/arweaveHashNEW/5", creatorCore1.uri(5));
+    vm.stopPrank();
+  }
+}
```
nikki-kiga, line 583: assumes if we're updating the tokenURI we would update for ALL of the tokens
and if the creator updates the image even for one they would need to reupload the whole series

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,257 @@
...
+   * See {IGachaLazyClaim-mintReserve}.
+   */
+  function mintReserve(address creatorContractAddress, uint256 instanceId, uint32 mintCount) external payable override {
+    if (Address.isContract(msg.sender)) revert IGachaLazyClaim.CannotMintFromContract();
+    Claim storage claim = _getClaim(creatorContractAddress, instanceId);
+    // Checks for reserving
+    if (claim.startDate > block.timestamp) revert IGachaLazyClaim.ClaimInactive();
+    if (claim.endDate > 0 && claim.endDate <= block.timestamp) revert IGachaLazyClaim.ClaimInactive();
+    if (msg.value != (claim.cost + MINT_FEE) * mintCount) revert IGachaLazyClaim.InvalidPayment();
+    uint32 amountAvailable = claim.totalMax - claim.total;
```
wwhchung, line 154: This seems wrong.  What if totalMax is zero (unlimited)

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
wwhchung: This is still incorrect @nikki-kiga 

You're not updating the storageProtocol with what's specified in he claimParameters

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,268 @@
...
+   * See {IGachaLazyClaim-mintReserve}.
+   */
+  function mintReserve(address creatorContractAddress, uint256 instanceId, uint32 mintCount) external payable override {
+    if (Address.isContract(msg.sender)) revert IGachaLazyClaim.CannotMintFromContract();
+    Claim storage claim = _getClaim(creatorContractAddress, instanceId);
+    // Checks for reserving
+    if (mintCount == 0 || mintCount >= MAX_UINT_32) revert IGachaLazyClaim.InvalidMintCount();
+    if (claim.startDate > block.timestamp || (claim.endDate > 0 && claim.endDate < block.timestamp)) revert IGachaLazyClaim.ClaimInactive();
+    if ((claim.totalMax != 0 && claim.total == claim.totalMax)) revert IGachaLazyClaim.ClaimSoldOut();
+    if (claim.total == MAX_UINT_32 || claim.total + mintCount > MAX_UINT_32) revert IGachaLazyClaim.TooManyRequested();
```
wwhchung, line 161: This can be simplified to:
if (claim.total + mintCount > MAX_UINT_32) revert IGachaLazyClaim.TooManyRequested();

The first part of the if statement is unecessary

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,268 @@
...
+    if (msg.value != (claim.cost + MINT_FEE) * mintCount) revert IGachaLazyClaim.InvalidPayment();
+    // calculate the amount to reserve and update totals
+    uint32 amountToReserve = mintCount;
+    if (claim.totalMax != 0) {
+      uint32 amountAvailable = claim.totalMax - claim.total;
+      amountToReserve = uint32(Math.min(mintCount, amountAvailable));
```
wwhchung, line 167: Why not one line this?
amountToReserve = uint32(Math.min(mintCount, claim.totalMax - claim.total));

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,268 @@
...
+  /**
+   * See {IGachaLazyClaim-mintReserve}.
+   */
+  function mintReserve(address creatorContractAddress, uint256 instanceId, uint32 mintCount) external payable override {
+    if (Address.isContract(msg.sender)) revert IGachaLazyClaim.CannotMintFromContract();
+    Claim storage claim = _getClaim(creatorContractAddress, instanceId);
+    // Checks for reserving
+    if (mintCount == 0 || mintCount >= MAX_UINT_32) revert IGachaLazyClaim.InvalidMintCount();
+    if (claim.startDate > block.timestamp || (claim.endDate > 0 && claim.endDate < block.timestamp)) revert IGachaLazyClaim.ClaimInactive();
+    if ((claim.totalMax != 0 && claim.total == claim.totalMax)) revert IGachaLazyClaim.ClaimSoldOut();
```
wwhchung, line 160: erroneous brackets
if (claim.totalMax != 0 && claim.total == claim.totalMax)

packages/manifold/contracts/gachaclaims/ERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,268 @@
...
+    if (msg.value != (claim.cost + MINT_FEE) * mintCount) revert IGachaLazyClaim.InvalidPayment();
+    // calculate the amount to reserve and update totals
+    uint32 amountToReserve = mintCount;
+    if (claim.totalMax != 0) {
+      uint32 amountAvailable = claim.totalMax - claim.total;
+      amountToReserve = uint32(Math.min(mintCount, amountAvailable));
+    }
+    claim.total += amountToReserve;
+    _mintDetailsPerWallet[creatorContractAddress][instanceId][msg.sender].reservedCount += amountToReserve;
+    if (amountToReserve > 0 && claim.cost > 0) {
```
wwhchung, line 171: amountToReserve HAS to be > 0, the first part of the if is unecessary

packages/manifold/contracts/gachaclaims/GachaLazyClaim.sol
```diff
@@ -0,0 +1,80 @@
...
+
+  string internal constant ARWEAVE_PREFIX = "https://arweave.net/";
+  string internal constant IPFS_PREFIX = "ipfs://";
+  address internal _signer;
+
+  uint256 internal constant MAX_UINT_8 = 0xff;
+  uint256 internal constant MAX_UINT_32 = 0xffffffff;
+  uint256 internal constant MAX_UINT_48 = 0xffffffffffff;
+  uint256 internal constant MAX_UINT_56 = 0xffffffffffffff;
+  uint256 internal constant MAX_UINT_80 = 0xffffffffffffffffff;
```
jaxonL, line 24: i think this one is wrong -- the actual value is MAX_UINT_72 instead of MAX_UINT_80
```suggestion
  uint256 internal constant MAX_UINT_80 = 0xffffffffffffffffffff;
```

packages/manifold/contracts/gachaclaims/IERC1155GachaLazyClaim.sol
```diff
@@ -0,0 +1,102 @@
...
+   * @param instanceId                the instanceId of the burnRedeem for the creator contract
+   * @param storageProtocol           the storage protocol for the metadata
+   * @param location                  the location of the metadata
+   */
+  function updateTokenURIParams(
+    address creatorContractAddress,
+    uint256 instanceId,
+    StorageProtocol storageProtocol,
+    string calldata location
+  ) external;
```
jaxonL, line 101: question: does this exist to easily swap the token uri, without needing to update the whole claim? since both update claim and update token uri have a way to update the location

packages/manifold/contracts/gachaclaims/IGachaLazyClaim.sol
```diff
@@ -0,0 +1,106 @@
...
+    address indexed creatorContract,
+    uint256 indexed instanceId,
+    address indexed collector,
+    uint32 mintCount
+  );
+
+  struct VariationMint {
+    uint8 variationIndex;
+    uint32 amount;
+    address recipient;
```
jaxonL, line 53: minor: if my understanding is correct -- should we be swapping the ordering for struct packing into 2 storage slots instead of 3 here?

eg:
```suggestion
    uint8 variationIndex;
    address recipient;
    uint32 amount;
```

jaxonL (changes requested):
very minor comment:
- do we want to rename mentions of "claim" to gacha?

blocking just on the constant value for now. will re-review erc1155 in more detail once we get the widget out ;)

packages/manifold/contracts/gachaclaims/IERC1155GachaLazyClaim.sol
nikki-kiga: yeah i think it's to be consistent with our other contracts but i think that's the core of the idea

PR #101, opened 2025-01-02
able to set mintFees for lazyClaim

donpdang:
## This PR

We want to be able to set Manifold mint fees dynamically. Use case is for polygon network, MATIC cost is different than ETH so we need to be able to accoun and set the fees however we like.
The changes are:
+ Add `setMintFees` to `LazyPayableClaim`. This is gated by `adminRequired` modifier
+ Add tests
+ Add ability to pause claim initializations

### Ticket

https://linear.app/manifoldxyz/issue/UPDATE-THIS-PART

## Self-Review

- [ ] Review and comment PR file changes
- [ ] Description provides context for the changes
- [ ] Tests are included for the PR (required on server changes)

## QA Steps

- [ ]


nikki-kiga (approved):
lgtm!
sidenote - is it possible there's an existing bug with insufficient funds check in _transferFunds that isn't inclusive of the mintfee?

wwhchung:
This is not a good PR.  We need to store the mint fee in the claim itself so we don't rug old claims when the fee changes.

wwhchung:
We need to store it efficiently in the claim.  How many bits do we have left?  Can we be smart and store in gwei instead of wei to reduce the amount of bits we need?

packages/manifold/contracts/lazyclaim/ERC1155LazyPayableClaim.sol
```diff
@@ -34,7 +34,8 @@ contract ERC1155LazyPayableClaim is IERC165, IERC1155LazyPayableClaim, ICreatorE
             interfaceId == type(IERC165).interfaceId;
     }
 
-    constructor(address initialOwner, address delegationRegistry, address delegationRegistryV2) LazyPayableClaim(initialOwner, delegationRegistry, delegationRegistryV2) {}
+    constructor(address initialOwner, address delegationRegistry, address delegationRegistryV2, uint256 mintFee,
```
wwhchung, line 37: Do not pass it into constructor.  Default to a value and then set.  Otherwise we will not have consistent addresses across chains with different fees

wwhchung (changes requested):
See all my comments.

packages/manifold/script/ERC1155LazyPayableClaim.s.sol
```diff
@@ -28,14 +28,24 @@ import "../contracts/lazyclaim/ERC1155LazyPayableClaim.sol";
         // address addr = vm.addr(pk);
         // console.log(addr);
 
+        // IMPORTANT: Set mint fees here
+        uint256 mintFee = 500000000000000;
```
wwhchung, line 32: Do not set fees in script

wwhchung:
1. We should not pass the fee to constructor, otherwise addresses will be different on different chains
2. If you are allowing fee changes to the contract, then you will need to store the current fee at time of claim initialization, and you will need to read the fee from the claim
3. If you are allowing for #2, then we need to store the fee as efficient as possible.  Maybe in gwei, maybe in some other thing.  But we shouldn't exceed the bits that would push the lazyclaim object to another storage slot
4. If 3 is not possible, maybe you ONLY allow the fee to be set once.

donpdang:
> lgtm! sidenote - is it possible there's an existing bug with insufficient funds check in _transferFunds that isn't inclusive of the mintfee?

hmm I thought `_transferFuns` is  inclusive of the mintFee according to 
https://github.com/manifoldxyz/creator-core-extensions-solidity/blob/e2be17dcc53d37104d418586cb36418860fef2a9/packages/manifold/contracts/lazyclaim/LazyPayableClaim.sol#L103

packages/manifold/contracts/lazyclaim/ILazyPayableClaim.sol
```diff
@@ -42,6 +42,16 @@ interface ILazyPayableClaim {
      */
     function setMembershipAddress(address membershipAddress) external;
 
+    /**
+    *  @notice Set the mint fee for non-merkle claims
+    */
+    function setMintFee(uint256 mintFee) external;
```
wwhchung, line 48: We should just have one function that takes in both.

donpdang: great point, just pushed the fix

wwhchung (approved):

